PR #7, opened 2023-11-18
Add mui to deps

naconnors:
## Description
Add mui to dependencies.

### Changes
* Add mui deps
* Remove tailwindcss dep

### Concerns
Assumes not using tailwindcss with mui


calebbourg (approved):

jhodapp (approved):
Verified locally, still builds cleanly with Node@21 and runs. Thanks Nick.

Issue #14, opened 2024-03-19
Select Default Organization and Coaching Relationship


jhodapp:
**Description:**
As a coach or coachee, I want to be able to set my default organization and coaching relationship, so that I can quickly access the relevant most coaching session details the next time I log in.

**Definition of Done:**
- I have a UI to set my default organization
- I have a UI to set my default coaching relationship
- I can set my default organization given my `user.external_id` via an API endpoint
- I can set my default coaching relationship given my `user.external_id` via an API endpoint

PR #18, opened 2024-06-10
17 fix auth store refresh page bug

qafui:
## Description
Updates the useEffect function to fix the infinite loop occurring on page refresh. 

### Changes
* Added userUUID to the dependency array of useEffect. This ensures the effect only runs when userUUID changes.
* Added a check in loadOrganizations to prevent fetching if userUUID is not available. This avoids unnecessary calls on initial render.
* Includes the changes by Jim from #14 making that PR no longer necessary.

### Testing Strategy
- Login with valid credentials
- Refresh the dashboard
- Confirm infinite loop is absent by checking the console for errors

### Concerns


qafui:
@Jim: decided to separate refactoring to include SWR & simplification of state management from the bug. IIt was less about Zustand loosing a stored value on refresh and more about passing the userUUID as a prop (rather than using state) AND leaving the dependency array empty.

jhodapp (approved):
Looks great, a nice simple fix. Thanks very much David - I learned something from this!

PR #19, opened 2024-06-14
Use store instead of props to pass userUUID

qafui:
## Description
Cleans up the `SelectCoachingSession` component to use the Zustand store for grabbing userUUID instead of props.
 
### Changes
* ...
* ...
* ...

### Testing Strategy

### Concerns

### DIFF
--- a/src/components/ui/dashboard/select-coaching-session.tsx
+++ b/src/components/ui/dashboard/select-coaching-session.tsx
@@ -22,15 +22,7 @@ import { useAuthStore } from "@/lib/providers/auth-store-provider";
 import { Organization, defaultOrganizations } from "@/types/organization";
 import { useEffect, useState } from "react";
 
-export interface CoachingSessionProps {
-  /** The current logged in user's UUID */
-  userUUID: string;
-}
-
-export function SelectCoachingSession({
-  userUUID,
-  ...props
-}: CoachingSessionProps) {
+export function SelectCoachingSession() {
   const [organizationSelection, setOrganizationSelection] =
     useState<string>("");
   const [relationshipSelection, setRelationshipSelection] =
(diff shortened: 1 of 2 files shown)


qafui:
@jhodapp Merged in your ux tweaks. We're good to 🚀

src/components/ui/dashboard/select-coaching-session.tsx
```diff
@@ -58,19 +51,18 @@ export function SelectCoachingSession({
   const [organizations, setOrganizations] = useState<Organization[]>(
     defaultOrganizations()
   );
-  useEffect(() => {
-    async function loadOrganizations() {
-      if (!userUUID) return;
+  const { userUUID } = useAuthStore((state) => state);
```
jhodapp, line 54: I still don't prefer to break encapsulation of the SelectionCoachingSession component by directly using `useAuthStore()` instead of passing it as a property one level deep. [Here's a good explanation of my thought process from Perplexity.ai](https://www.perplexity.ai/search/is-it-proper-J6l4Nqz3TECfYyL6DC9I8A) summarizing other sites on this.

Here's my own thought process: I think it's a logical practice to retrieve state at the page level and pass user session state like userUUID down to components because components _are_ reusable and pages are not. But a component like this one becomes less reusable when tightly coupling the AuthStore into the innards of the component.

jhodapp (changes requested):
Thanks for this David, I left my latest thoughts inline. Let me know what you think.

I know you mentioned [this article from Vercel](https://vercel.com/guides/react-context-state-management-nextjs) on passing state via Context, but I don't think their example applies directly in this case.

calebbourg (approved):
I was able to test this successfully against https://github.com/Jim-Hodapp-Coaching/refactor-platform-rs/pull/50
Screen Shot 2024-07-02 at 8 54 45 AM

Awesome work!

PR #23, opened 2024-07-05
Keep track of organizationId, relationshipId and coachingSessionId across routes

jhodapp:
## Description
Implements a Zustand store to keep track of organizationId, relationshipId and coachingSessionId across multiple app routes.

### Changes
* Create a new Zustand store that's available to all App routes called app-state-store
* Use this store in the select-coaching-session component
* Clean up some unused console statements and downgrade some to trace level

N/A

### Testing Strategy
1. Everything should continue to work as it did before in terms of selecting an organization, relationship and coaching session
2. The only difference is you should see the select-coaching-session component select boxes retain their selected values if you join a session and return to the dashboard page without logging out
3. You should also be able to observe that these values get reset once you log out of the user session

### Concerns
None


qafui (approved):
I like the simplified ID names.

jhodapp:
> I like the simplified ID names. 

Thanks, I do too. Makes things name and type consistent and simpler.

PR #24, opened 2024-07-07
Add SWR and useRequest hook

qafui:
## Description
This PR consolidates data fetching and error handling with the SWR library.

It allows for increased focus on implementing components without kitting out API fetchers each time. In addition, contributors will more easily implement loading, loaded and error states for each component with a uniform pattern that minimizes code duplication.

### Changes
* ...
* ...
* ...

### Testing Strategy

### Concerns


qafui:
TODO: Create a separate custom hook named useCoachingData that fetches the data using useRequest. Call useCoachingData in the SelectCoachingSession to fix the build error.

PR #25, opened 2024-07-25
Introduce MVP-level coaching session Note creation/updating

jhodapp:
## Description
This PR creates a new note when a coaching session page first loads and an associated note.id isn't yet set. It also updates the singular Note (for now) associated with this session after a 1 second delay of non-typing.

**Note:** this requires the use of this [backend PR](https://github.com/Jim-Hodapp-Coaching/refactor-platform-rs/pull/57) to function correctly

### Changes
* Introduces a new `Note` type
* Create a new Note instance if there isn't one yet for the given `coaching_session_id`
* Fetches the Note's body on page render if there is an instance already associated with the given `coaching_session_id`
* Updates the Note's body on the backend if the user types any changes

### Testing Strategy
1. Join a coaching session
2. Observe the Note body text in the textarea if there is already Note text associated with this session
3. Update the text, notice that a text indicator below the note textarea says "All change saved" if it was able to update the backend, otherwise should say "Failed to save changes" if it couldn't update the backend

### Concerns
This PR does not yet automatically update the second user's Note textarea when the first user makes any changes. I haven't tested this across multiple users yet, they will most likely clobber each other's updates if they both make Note changes.

### DIFF
--- a/src/app/coaching-sessions/[id]/page.tsx
+++ b/src/app/coaching-sessions/[id]/page.tsx
@@ -40,9 +40,14 @@ import { models, types } from "@/data/models";
 import { current, future, past } from "@/data/presets";
 import { useAppStateStore } from "@/lib/providers/app-state-store-provider";
 import { useEffect, useState } from "react";
-import { createNote } from "@/lib/api/notes";
-import { noteToString } from "@/types/note";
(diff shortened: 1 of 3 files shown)


calebbourg (approved):
I was able to test this locally. Very cool!!

Screen Shot 2024-08-01 at 6 56 54 AM

PR #27, opened 2024-08-17
A few UI/UX tweaks to the coaching session page

jhodapp:
## Description
This PR makes a few UI tweaks to prepare the page layout for adding the main missing elements like agreements, actions, etc.

### Changes
* Fix the overarching goal div alignment to always line up with the coaching notes div on the left and right sides
* Ensure the coaching session title div stays aligned with coaching notes on left and right sides as the window size changes
* Remove the CodeViewer and PresetShare components from the coaching session page

Screenshot 2024-08-20 at 21 54 04

### Testing Strategy
1. Navigate to the coaching sessions page
2. Change the window size from very wide to very narrow and notice the above alignment tweaks hold true

### Concerns
None

### DIFF
--- a/src/app/coaching-sessions/[id]/page.tsx
+++ b/src/app/coaching-sessions/[id]/page.tsx
@@ -129,7 +129,7 @@ export default function CoachingSessionsPage() {
   return (
     <>
       <div className="hidden h-full flex-col md:flex">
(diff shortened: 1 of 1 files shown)


qafui (approved):
Looks good. 👍🏾

PR #34, opened 2024-10-02
Refactor Select Coaching Session Component

qafui:
## Description

This PR adds the `useSWR` hook from the `swr` library as a wrapper around a global fetcher hook (useApiData). It allows any component to access the data it needs for rendering dynamic values from the API.

A dashboard component is refactored to use the new hook to fetch data for conditionally rendered select components and a button. The `JoinCoachingSession` component allows a user to select an organization, a coaching relationship and a session and join the session with a button click. 

#### Todo

- [x] Separate previous from upcoming sessions with a `SelectGroup`
- [x] Merge and fix conflicts

### Changes
* Adds a hook for fetching data from the API
* Adds a Dynamic select component that renders dropdowns with data fetched from the API
* Renders 3 dynamic select components in a composite or ‘smart’ `JoinCoachingSession` component responsible for setting organization, relationship and session IDs in the state tree.
Screenshot 2024-10-03 at 3 12 08 PM

### Testing Strategy
Login and attempt to join a session using the component illustrated in the image

### Concerns
- The rendering helper functions in the `dynamic-api-select` component should be separated into its own component.
- PRs of this kind should be broken up into smaller features in future to improve code merge 😅


src/hooks/use-api-data.ts
```diff
@@ -0,0 +1,60 @@
+import useSWR from 'swr';
+import { siteConfig } from '@/site.config';
+
+interface FetcherOptions {
+    url: string,
+    method?: 'GET' | 'POST',
+    params?: Record<string, string>,
+    // body?: Record<string, any>
```
jhodapp, line 8: @qafui is this intentional to be commented out?

qafui: Yes. I realized that this hook is a little too generic. I suggest we split it into a few custom hooks (one for credentialed `POST` fetches and another for simple `GET` requests where we can use more `Cache` and revalidation config options to change the behavior of the `useSWR` hook. We would need that for taking notes and more live streaming behaviors.

PR #41, opened 2024-10-10
Add a new type called UserSession and use it in the UserNav component

jhodapp:
## Description
This PR adds a new type called UserSession and uses it in the UserNav component.

### Changes
* Add a new type called UserSession which holds more than just userId
* Use the new fields from the AuthStore in the UserNav component

Refactor Coaching _ Mentoring

### Testing Strategy
1. Log in
2. View the UserNav menu (upper right hand corner menu)
3. Notice that the 2 letter icon is dynamically set to "FL" where FL == [F]irst + [L]ast
4. Notice that the user's full name is dynamically set
5. Notice that the user's email address is dynamically set

### Concerns
None


jhodapp (commented):
Self-approving while David's fully occupied this week.

PR #67, opened 2025-02-06
Split apart choosing relationship and coaching sessions on the Dashboard

calebbourg:
## Description

This PR updates the Dashboard page in order to allow for creating new coaching sessions.

**Relates to:** https://github.com/refactor-group/refactor-platform-rs/pull/99

### Changes
* Splits choosing Organizations/Coaching Relationships apart from how Coaching Sessions are presented
* Renders a list of Coaching Sessions instead of a dropdown
* Adds a "Join Session" button to each Coaching Session
* Adds a "Create Session" Button and modal for creating a new Coaching Session

#### Successful Response from the BE with newly created Coaching Session and `collab_document_name`
Screenshot 2025-02-11 at 7 18 54 AM

New Tiptap document created as part of Coaching Session creation
Screenshot 2025-02-11 at 7 19 08 AM

### Concerns
The addition of a modal for creating a new coaching session may not be what we want long term. I added it here like this because I wasn't sure where to place the form.

### Coming Next
- Potentially allowing a user to set an Overarching Goal on Coaching Session creation


.env
jhodapp, line 1: Just an FYI, we don't want to commit this file with uncommented environment variables because these get set by the backend docker-compose.yaml for production builds.

jhodapp: It may make sense to add this to .gitignore

calebbourg: Good call! I removed that from the commit

src/components/ui/dashboard/coaching-session-list.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const [sortByDate, setSortByDate] = useState(true);
+  const [isDialogOpen, setIsDialogOpen] = useState(false);
+  const [newSessionDate, setNewSessionDate] = useState("");
+
+  const handleCreateSession = (e: React.FormEvent) => {
+    e.preventDefault();
+
+    // Format the date string
+    const formattedDate = format(
+      new Date(newSessionDate),
```
jhodapp, line 46: I've been using ts-luxmon for working with Dates and Times because it's far more powerful and flexible in its parsing of standard date/time-stamps (e.g. ISO standards).

Do you agree that consistency is important? If so, take a look at how I use ts-luxmon in src/types/session-title.ts as an example. I also allow for one's locale to influence the formatting of the date with a system-wide config value.

src/components/ui/dashboard/coaching-session-list.tsx
```diff
@@ -0,0 +1,154 @@
...
+          </CardTitle>
+          <Dialog open={isDialogOpen} onOpenChange={setIsDialogOpen}>
+            <DialogTrigger asChild>
+              <Button
+                variant="outline"
+                size="sm"
+                className="w-full sm:w-auto"
+                disabled={!isCoach || !currentCoachingRelationshipId}
+              >
+                <CalendarPlus className="mr-2 h-4 w-4" />
```
jhodapp, line 89: We already have a Calendar component from shadcn to utilize which matches the style shadcn brings. See my use of it here: https://github.com/refactor-group/refactor-platform-fe/blob/a0f620610517252c6d456513a7b844b55927fac8/src/components/ui/coaching-sessions/actions-list.tsx#L335

It does currently have a few issues (e.g. I can't seem to get the current date to highlight, nor the currently-selected date when updating an action's existing due by date) since I upgraded to React 15 and everything that brings with it, but in general the aim is to use shadcn components over Lucid ones where possible.

Perhaps there's a newer version of the Calendar component that came out more recently than the last time I was working with this code, which was mid to late December. You can use [shadcn's CLI](https://ui.shadcn.com/docs/cli) to pull a new version of the component into our repo.

[Shadcn's Calendar documentation](https://ui.shadcn.com/docs/components/calendar)

src/lib/api/user-session.ts
```diff
@@ -40,6 +40,13 @@ export const loginUser = async (
       const userSessionData = response.data.data;
       if (isUserSession(userSessionData)) {
         userSession = parseUserSession(userSessionData);
+      } else {
+        // Even if the log in is successful, if we don't have user data that we can
+        // deserialize successfully, then downstream operations will see the default
+        // userSessionData in state and we will experience subtle Bugs. We should consider
+        // how best we want to handle this. Ex. clear auth cookie?
+        console.debug("userSessionData: ", userSessionData)
+        throw { message: 'Login Failed to Produce valid User Session Data' }
```
jhodapp, line 49:
```suggestion
        throw { message: 'Login Failed to produce valid User Session data' }
```

I use a pattern of capitalizing only the beginning of a sentence and proper nouns (e.g. entities/objects), otherwise everything else is lowercase. I imagine it seems a little pedantic, but I try and strive for this level of consistency.

src/lib/stores/auth-store.ts
```diff
@@ -36,6 +40,13 @@ export const createAuthStore = (initState: AuthState = defaultInitState) => {
           logout: () => {
             set(defaultInitState);
           },
+          setIsCoach: (coachId) => {
+            var userId = get().userId;
+            set({ isCoach: !!(userId && coachId && userId == coachId) });
```
jhodapp, line 45: I think it still makes sense to store `isCoach` as a property in this store, but perhaps it makes sense to move this logic into a function as part of the UserSession type, in src/types/user-session.ts? That seems to me to be grouping like things together where one would expect them. Let me know what you think.

jhodapp (changes requested):
Thanks Caleb, really strong first frontend PR! Some inline suggestions and questions for you.

src/components/ui/dashboard/coaching-session-list.tsx
calebbourg: Updated to use the shadcn component in 6b44c1f0889e67902721cd6ac23da4d95d278c06. Thanks for the heads up on that!

src/components/ui/dashboard/coaching-session-list.tsx
calebbourg: Completely agree efe7867e005c8b1b49019cd7b53dcb76d9d710a9. I blindly went with what v0.dev offered :)

src/components/ui/coaching-session.tsx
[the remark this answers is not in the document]
calebbourg: I _think_ I fixed this in 3be4f294a59df3fad0379e0c0fe242015a0a4174

src/lib/stores/auth-store.ts
calebbourg: Can you elaborate on your thinking around this? In my mind the logic belongs here because the `isCoach` concept is a part of the `AuthState` interface because it's a derived virtual state that is dependent on the coaching relationship context as opposed to a more static property of the `UserSession` (Ex. `first_name`).

I want to make sure I understand where you're coming from though as this is still fairly new to me.

src/components/ui/dashboard/coaching-session-list.tsx
calebbourg: I agree that the modal and the general UX is not ideal here. My thoughts are to push through with the plumbing and make sure you _can_ create a new Coaching Session and then we can come back and make things look and feel pretty.

src/components/ui/coaching-session.tsx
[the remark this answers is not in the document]
jhodapp: Much better, thank you!

src/components/ui/dashboard/coaching-session-list.tsx
```diff
@@ -42,8 +42,7 @@ export default function CoachingSessionList() {
     e.preventDefault();
 
     // Format the date string
-    const formattedDate = format(
-      new Date(newSessionDate),
+    const formattedDate = DateTime.fromISO(newSessionDate).toFormat(
```
jhodapp: I created a nice [function to work with ISO DateTime strings here](https://github.com/refactor-group/refactor-platform-fe/blob/a0f620610517252c6d456513a7b844b55927fac8/src/types/session-title.ts#L71C5-L73C46) that I recommend using:

```typescript
const formattedDate = getDateTimeFromString(newSessionDate)
      .setLocale(locale)
      .toLocaleString(DateTime.DATETIME_SHORT);
```

More about [ts-luxon's formatting functions here](https://tonysamperi.github.io/ts-luxon/docs/#/formatting?id=tolocalestring-strings-for-humans).

src/components/ui/calendar.tsx
```diff
@@ -40,22 +24,22 @@ function Calendar({
...
-          "flex-1 text-center text-muted-foreground rounded-md w-9 font-normal text-[0.8rem]",
+          "text-muted-foreground rounded-md w-9 font-normal text-[0.8rem]",
         row: "flex w-full mt-2",
-        day: "flex-1 text-center text-sm p-0 relative [&:has([aria-selected].day-range-end)]:rounded-r-md [&:has([aria-selected].day-outside)]:bg-accent/50 [&:has([aria-selected])]:bg-accent first:[&:has([aria-selected])]:rounded-l-md last:[&:has([aria-selected])]:rounded-r-md focus-within:relative focus-within:z-20",
-        day_button: cn(
+        cell: "h-9 w-9 text-center text-sm p-0 relative [&:has([aria-selected].day-range-end)]:rounded-r-md [&:has([aria-selected].day-outside)]:bg-accent/50 [&:has([aria-selected])]:bg-accent first:[&:has([aria-selected])]:rounded-l-md last:[&:has([aria-selected])]:rounded-r-md focus-within:relative focus-within:z-20",
+        day: cn(
           buttonVariants({ variant: "ghost" }),
-          "h-9 w-9 p-0 font-normal aria-selected:opacity-100 mx-auto"
+          "h-9 w-9 p-0 font-normal aria-selected:opacity-100"
```
jhodapp, line 42: I had these custom changes in here because the Calendar component doesn't quite render correctly when using React@19 and the other dependencies it relies on.

Screenshot 2025-02-16 at 12 19 48

See what I mean? The day labels above the calendar aren't lined up, the <> arrows aren't where they should be, etc.

We can talk about the best way to handle this if you're unsure.

jhodapp (changes requested):

src/lib/stores/auth-store.ts
jhodapp: Yeah I think it's fine for now, I was over-thinking this one.

src/components/ui/calendar.tsx
calebbourg: oh I see that yeah. These were updated automatically when I updated the calendar component using the CLI. Let's chat about it yeah

jhodapp: If you leave the style changes in place and the temporary comment that I put in there, it should fix it well enough to move on. There's more we should do, and probably need to reach out to shadcn about this, but it's what I could figure out when working on this problem around Christmas time.

src/components/ui/dashboard/coaching-session-list.tsx
calebbourg: Updated! f8d2a81db08ab15150c0b075293169be70fcca3a

jhodapp (approved):
LGTM

PR #68, opened 2025-02-20
Add tiptap

calebbourg:
## Description
_describe the intent of your changes here_

### Changes
* ...
* ...
* ...

### Testing Strategy

### Concerns

### DIFF
--- a/src/app/coaching-sessions/[id]/page.tsx
+++ b/src/app/coaching-sessions/[id]/page.tsx
@@ -18,22 +18,19 @@ import { useAuthStore } from "@/lib/providers/auth-store-provider";
 import { siteConfig } from "@/site.config";
 import { CoachingSessionTitle } from "@/components/ui/coaching-sessions/coaching-session-title";
 import { OverarchingGoalContainer } from "@/components/ui/coaching-sessions/overarching-goal-container";
-import {
-  HoverCard,
-  HoverCardContent,
-  HoverCardTrigger,
-} from "@/components/ui/hover-card";
-import { Label } from "@/components/ui/label";
-import { Button } from "@/components/ui/button";
+import { TipTapEditor } from "@/components/ui/coaching-sessions/tiptap-editor";
+// import {
+//   HoverCard,
+//   HoverCardContent,
+//   HoverCardTrigger,
+// } from "@/components/ui/hover-card";
+// import { Label } from "@/components/ui/label";
+// import { Button } from "@/components/ui/button";
 import { LockClosedIcon, SymbolIcon } from "@radix-ui/react-icons";
-import {
-  CoachingNotes,
-  EditorRef,
-} from "@/components/ui/coaching-sessions/coaching-notes";
 import CoachingSessionSelector from "@/components/ui/coaching-session-selector";
 import { useRouter } from "next/navigation";
 import { useCoachingRelationshipStateStore } from "@/lib/providers/coaching-relationship-state-store-provider";
-import { useCoachingSessionStateStore } from "@/lib/providers/coaching-session-state-store-provider";
+// import { useCoachingSessionStateStore } from "@/lib/providers/coaching-session-state-store-provider";
 
 // TODO: can't make this page be server side yet until we move useRouter/useEffect out of this page
 // export const metadata: Metadata = {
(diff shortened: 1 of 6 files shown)


src/site.config.ts
```diff
@@ -22,6 +22,7 @@ export const siteConfig = {
     backendApiVersion: process.env.NEXT_PUBLIC_BACKEND_API_VERSION,
     frontendServicePort: process.env.FRONTEND_SERVICE_PORT,
     frontendServiceInterface: process.env.FRONTEND_SERVICE_INTERFACE,
+    tiptapAppId: process.env.NEXT_PUBLIC_TIPTAP_APP_ID,
```
jhodapp, line 25: @calebbourg Should this be build time scope or public runtime scope? Just want to double check since it's a sensitive variable.

https://nextjs.org/docs/app/building-your-application/deploying#environment-variables

src/components/ui/coaching-sessions/tiptap-editor.tsx
```diff
@@ -1,266 +1,94 @@
...
+        name: jwt.sub,
+        appId: tiptapAppId,
+        token: jwt.token,
+        document: doc,
+        user: userSession.display_name,
+        connect: true,
+        broadcast: true,
+      });
+
+      setProvider(newProvider);
```
zgavin1, line 50: I'm still setting getting branches running up so I could just be totally wrong, but we may not need the stateful reference to the provider. Will test this theory if when I get it up and running and confirm or correct myself.

calebbourg: Yeah you're probably totally right. This is one of those AI decisions that I wasn't sure about but went with, blindly, because it seemed to work. I'd be very interested to get your thoughts on how this should be set up.

src/site.config.ts
calebbourg: @jhodapp what's our pattern for having this injected at build time?

jhodapp: The other variables that don't have `NEXT_PUBLIC_*` in front of them follow this pattern. You can see how they get injected in our [docker-compose.yaml file here](https://github.com/refactor-group/refactor-platform-rs/blob/8852758ce3dd5c95950b5aaf6b236072a1b24940/docker-compose.yaml#L64) from the backend. But if you set them directly here in the .env, they'll be part of the environment, just not exposed to Next.js.

src/components/ui/coaching-sessions/tiptap-editor.tsx
jhodapp: And some more [here](https://github.com/ueberdosis/hocuspocus/blob/main/docs/provider/examples.md) too (examples).

src/lib/api/collaborationToken.ts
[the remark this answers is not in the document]
calebbourg: Updated in [d048137](https://github.com/refactor-group/refactor-platform-fe/pull/68/commits/d04813764cb45024fbab8a8a13343cbdd7c1c7bf)

src/site.config.ts
calebbourg: updated in d04813764cb45024fbab8a8a13343cbdd7c1c7bf

zgavin1 (approved):

README.md
```diff
@@ -14,23 +14,32 @@ This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next
...
+When running locally on a development machine you can manually set the application's configuration through a `.env` file at the root of the source tree:
+
+```env
+NEXT_PUBLIC_BACKEND_SERVICE_PROTOCOL="http"
+NEXT_PUBLIC_BACKEND_SERVICE_PORT=4000
+NEXT_PUBLIC_BACKEND_SERVICE_HOST="localhost"
+NEXT_PUBLIC_BACKEND_API_VERSION="0.0.1"
+
+# TIPTAP_APP_ID originates from your TipTap Cloud Dashboard
+NEXT_PUBLIC_TIPTAP_APP_ID="<TIPTAP_APP_ID>"
```
jhodapp, line 28:
```suggestion
TIPTAP_APP_ID="<TIPTAP_APP_ID>"
```

src/app/coaching-sessions/[id]/page.tsx
```diff
@@ -183,13 +74,7 @@ export default function CoachingSessionsPage() {
...
               <TabsContent value="notes">
                 <div className="flex-col h-full space-y-4">
-                  <CoachingNotes
-                    ref={editorRef}
-                    value={note.body}
-                    onChange={handleOnChange}
-                    onKeyDown={handleKeyDown}
-                  ></CoachingNotes>
-                  <p className="text-sm text-muted-foreground">{syncStatus}</p>
+                  <TipTapEditor />
```
Comment (jhodapp) [inline src/app/coaching-sessions/[id]/page.tsx:77]: I originally had this component called CoachingNotes as a specific design decision, to keep the component used on the page as a generically named one (abstract) that would then instantiate the specific editor under it (concrete) called TipTapEditor.

My reasoning for this is so we can keep the level of tight coupling to TipTap as low as possible. Would you be open to restoring this pattern?

src/components/ui/coaching-sessions/coaching-notes.tsx
jhodapp, line 1: Do we have any need for `/src/lib/api/coaching-notes.ts` `/src/lib/types/note.ts` anymore on the frontend? Or does it still have a "virtual" role in some way used by the backend but has changed in the specific fields we need to store?

src/app/coaching-sessions/[id]/page.tsx
```diff
@@ -1,137 +1,28 @@
...
-// TODO: can't make this page be server side yet until we move useRouter/useEffect out of this page
-// export const metadata: Metadata = {
-//   title: "Coaching Session",
-//   description: "Main coaching session page",
-// };
 
 export default function CoachingSessionsPage() {
   const router = useRouter();
-  const [note, setNote] = useState<Note>(defaultNote());
-  const [syncStatus, setSyncStatus] = useState<string>("");
```
Comment (jhodapp) [inline src/app/coaching-sessions/[id]/page.tsx:47]: What do you think about still using `syncStatus` and connecting it to the TipTapEditor's document sync status? It'd be nice to know when some event statuses occur like:
1. Document initially synced (i.e. loaded)
2. Connected to TipTap cloud
3. Offline from TipTap cloud
4. Offline changes pending and not synced to cloud

Comment (jhodapp) [reply src/app/coaching-sessions/[id]/page.tsx:47]: https://github.com/refactor-group/refactor-platform-fe/blob/c0014b0f9f6f38acb5b73eb5a28b3b1ea7a8057e/src/app/coaching-sessions/%5Bid%5D/page.tsx#L69

src/components/ui/coaching-sessions/tiptap-editor.tsx
```diff
@@ -1,266 +1,94 @@
...
+  const { currentCoachingSessionId } = useCoachingSessionStateStore(
+    (state) => state
+  );
+  const { userSession } = useAuthStore((state) => ({
+    userSession: state.userSession,
+  }));
+  const { jwt, isLoading, isError } = useCollaborationToken(
+    currentCoachingSessionId
+  );
+  const [isSyncing, setIsSyncing] = useState(true);
```
jhodapp, line 25: Per my previous comment above, it'd be nice to surface this editor state up to the status label that sits under the editor's text box on the CoachingSession page.

src/site.config.ts
jhodapp:
```suggestion
    tiptapAppId: process.env.TIPTAP_APP_ID,
```

jhodapp (changes requested):

src/components/ui/coaching-sessions/coaching-notes.tsx
calebbourg: I had removed the API before. Removed the type file in 22fe9550b28afce40e219a7479bfc538d04cc6e8

src/app/coaching-sessions/[id]/page.tsx
Comment (calebbourg) [reply src/app/coaching-sessions/[id]/page.tsx]: Yup! very onboard with this 
2ce66d7266cbbd82f8c34b65e999e51be7acc5a5

README.md
calebbourg: I played around with this a bit and realized that the reason it wasn't working for us when we were pairing is that a value that **is not** prefixed with `NEXT_PUBLIC_` is not available to the browser at all and is only available for server rendered components. In our case we need it in a component rendered on the client. I _think_ this is ok, at least at the moment. Looking at the devtools we can see the App ID as part of the URL our tiptap library is using for the WS messaging.  So it's exposed via the internals of Tiptap itself.

src/app/coaching-sessions/[id]/page.tsx
Comment (calebbourg) [reply src/app/coaching-sessions/[id]/page.tsx:47]: I think this is a really great idea and 100% worth doing. Do you mind if we punt on it for the moment and maybe added to our list of cosmetic follow ups? I'd like to get this PR and the cursor PR landed so we can move on to the user/coaching session updates given our timeframe.

Comment (jhodapp) [reply src/app/coaching-sessions/[id]/page.tsx:47]: I'm ok with punting this into the small changes list.

README.md
jhodapp: Thanks for looking into this Caleb. If it's exposed anyway by normal TipTap operation, then I agree that it's ok to be a NEXT_PUBLIC_* variable.

calebbourg:
Thank you for all of the help and expertise here @zach and @jim

PR #70, opened 2025-02-27
Add tiptap with EditorProvider

zgavin1:
## Description

Use `EditorProvider` rather than `useEditor` makes the editor object available to all its children (such as the toolbar) via a context hook `useCurrentEditor`: https://tiptap.dev/docs/editor/getting-started/install/react#integrate-tiptap. The docs have examples of both methods. The provider pattern is calling `useEditor` to get the content anyway, so this pattern is the same except we can now access the editor in toolbar without drilling it as a prop.

Also add the `CollaborationProvider` "syncing" step to our "loading" state for the coaching notes so we don't see a blank text area before we see it loaded with the collab doc content: https://tiptap.dev/docs/collaboration/provider/events#display-connection-status

Minor adjustment to the collaboration token fetcher and query hook. No-op change, just readability.

https://www.loom.com/share/d394d0ecb7e74030807d1fc8bbc27fad

### Testing Strategy
Local dev testing

### DIFF
--- a/src/components/ui/coaching-sessions/tiptap-editor.tsx
+++ b/src/components/ui/coaching-sessions/tiptap-editor.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useEditor, EditorContent } from "@tiptap/react";
+import { EditorProvider } from "@tiptap/react";
 import { TiptapCollabProvider } from "@hocuspocus/provider";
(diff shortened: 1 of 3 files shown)


calebbourg (approved):
@zgavin1 This is great thank you!!

PR #71, opened 2025-02-28
Add tiptap cursor collab

calebbourg:
## Description
_describe the intent of your changes here_

### Changes
* ...
* ...
* ...

### Testing Strategy

### Concerns


src/components/ui/coaching-sessions/coaching-notes.tsx
```diff
@@ -46,8 +52,33 @@ const useCollaborationProvider = (doc: Y.Doc) => {
         },
       });
 
+      console.log("newProvider", newProvider);
+
       newProvider.on("synced", () => {
+        console.log("synced");
         setIsSyncing(false);
+        // Setting these here with the goal of only initializing things once
+        setProvider(newProvider);
```
calebbourg, line 61: I'm hoping that this is a reasonable approach. 

The hypothesis is that when we initialize multiple providers, it provides multiple instances of whatever is handling synchronization in the Tiptap cloud and so our browser editors are not syncing and behaving like we hoped they would.

Interestingly enough, I can see in my console that the provider instance being passed around holds a record of many clients

Screenshot 2025-03-03 at 8 43 39 AM

Presumably from me refreshing the page multiple times?

This has me pretty stumped but I think the screenshot is definitely a clue as to what might be going on.

jhodapp:
Replacing with #77 since this branch got very confused merging sub-branches and merging with main.

PR #75, opened 2025-03-10
Move login UI to index route & temp remove GitHub Oauth

zgavin1:
## Description
Move the existing login page UI to the index (/) route.

### Changes
- The existing index page was just boilerplate from NextJS, so it can be replaced entirely.
- The existing `/login/page.tsx` file (and thus the `/login` app route) can be removed (we could also re-include this and have it redirect to the index route)
- Comment out the GitHub oauth elements from the JSX

@jhodapp mentioned using [Route Groups](https://nextjs.org/docs/app/building-your-application/routing/route-groups). Reading the docs my understanding of that feature is it has a kinda different purpose. Happy to discuss more as I might be misreading

### Testing Strategy
1. Visit / and notice it's the same login form as what used to be at /login
2. Logout and notice that you're returned to / and the login form

### Concerns
I may want to try implementing some redirects before merging this. Leaving draft status for now.


src/app/page.tsx
zgavin1, line 1: This is a direct copy of the `login/page.tsx` file that was removed

jhodapp (approved):
This looks great Zach, I was thinking we might want a virtual directory for `login` but since there's currently nothing of substance at / anyway, your solution makes even better sense.

If we ever want to change what's at / we can just do so then. Thanks for this!

PR #76, opened 2025-03-12
Add a middleware layer, handle optimistic auth check on basic routes

zgavin1:
## Description
NextJS has a [cool middleware](https://nextjs.org/docs/app/building-your-application/routing/middleware) feature we can leverage.

If I'm logged in and try to visit the login index page, redirect me to the dashboard. If I'm logged out and I try to access /dashboard or /coaching-session/... route, redirect me to the login index page.

This is not meant to be an all-encompassing route protection system. Just a simple "optimistic'  light weight check against the cookie set by our login/logout API.

This does log a user out if their session expires from the backend.

### Testing Strategy

### Concerns
Exposes bug #79 by the use of localStorage but not responsible for the bug.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,0 +1,42 @@
+import { NextRequest, NextResponse } from "next/server";
+
+// 1. Specify protected and public routes
+const protectedRoutes = [
+  "/dashboard",
+  "/coaching-sessions",
+  "/settings",
+  "/profile",
+];
+const publicRoutes = ["/"];
+
+export default async function middleware(req: NextRequest) {
+  // 2. Check if the current route is protected or public
+  const path = req.nextUrl.pathname;
+  const isProtectedRoute = protectedRoutes.includes(path);
+  const isPublicRoute = publicRoutes.includes(path);
+
+  // 3. Decrypt the session from the cookie
+  const sessionCookie = req.cookies.get("id");
+  let session = sessionCookie?.value;
+
+  // 4. Redirect to / if the user is not authenticated
+  if (isProtectedRoute && !session) {
+    return NextResponse.redirect(new URL("/", req.nextUrl));
+  }
+
+  // 5. Redirect to /dashboard if the user is authenticated
+  if (
+    isPublicRoute &&
+    session &&
+    !req.nextUrl.pathname.startsWith("/dashboard")
+  ) {
+    return NextResponse.redirect(new URL("/dashboard", req.nextUrl));
+  }
+
+  return NextResponse.next();
+}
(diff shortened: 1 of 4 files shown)


src/middleware.ts
```diff
@@ -0,0 +1,42 @@
+import { NextRequest, NextResponse } from "next/server";
+
+// 1. Specify protected and public routes
+const protectedRoutes = [
+  "/dashboard",
+  "/coaching-sessions",
```
jhodapp, line 6: @zgavin1 I'm still able to visit /coaching-sessions/[id] when logged out. Are you seeing this too?

src/middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export default async function middleware(req: NextRequest) {
+  // 2. Check if the current route is protected or public
+  const path = req.nextUrl.pathname;
+  const isProtectedRoute = protectedRoutes.includes(path);
+  const isPublicRoute = publicRoutes.includes(path);
+
+  // 3. Decrypt the session from the cookie
+  const sessionCookie = req.cookies.get("id");
+  let session = sessionCookie?.value;
```
jhodapp, line 20:
```suggestion
  const session = sessionCookie?.value;
```

src/middleware.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  // 4. Redirect to / if the user is not authenticated
+  if (isProtectedRoute && !session) {
+    return NextResponse.redirect(new URL("/", req.nextUrl));
+  }
+
+  // 5. Redirect to /dashboard if the user is authenticated
```
jhodapp, line 27: I don't quite understand what this check/redirect are doing. Can you describe an example scenario?

jhodapp (changes requested):
Looks great Zach, thank you. A few things line to have a look at.

src/middleware.ts
zgavin1: yep good catch

src/middleware.ts
zgavin1: If I'm already logged in (have valid session cookie) and I navigate to "/" (the login page), redirect me to the dashboard. It's like the inverse of the first check.

jhodapp (approved):
Looks great, works as expected.

There is a newly exposed bug from moving to localStorage where the coaching_session_id contained in the URL vs the one in the AuthStore are used differently on the coaching session page and becomes evident when you try and open the same coaching session URL on a new tab of the same browser. But this isn't your fault, it's just a bug outside the scope of this PR.

PR #82, opened 2025-03-18
Set error state with message rather than with error object

### LINKED ISSUE #81: [Bug]: Handle auth login failure gracefully
### Contact Details

_No response_

### What happened?

Right now, if you enter invalid login credentials and try to submit the form, the page fully errors out. We should just be displaying an error message and letting the user try again.

### What is the version of refactor-platform-fe that you're seeing the problem on?

1

### Any related code or configuration details to produce this issue as well as relevant log output.

```shell

```

zgavin1:
## Description
Auth login failure was updating local error state for the auth component but we were setting the state value to the whole error object rather than the error message.

### Changes
* ...

### Testing Strategy
* Submit invalid login credentials
* Page does not whitescreen (aka show nextjs error page). login form should just gracefully display an error message.

### Concerns


jhodapp (approved):
Thanks for figuring this one out Zach.

Issue #80, opened 2025-03-18
[Refactor]: POST/PUT actions shouldn't send unnecessary entity fields


jhodapp:
### Contact Details

james.hodapp@gmail.com

### What happened?

Currently when calling a create/update action on any of the domain entities, we include all of the entity's data fields, including the ones that only ever get set/updated by the backend (e.g. `created_at` and `updated_at`).

We should only be passing along the fields that are set/updated by the frontend and required by the backend for successful create/update actions.

I believe this can be achieved pretty straightforwardly by changing all of the entity's `default*()` functions to only include these required fields. I don't believe we use these fields for any other operations other than create/update anyway.

### What is the version of refactor-platform-fe that you're seeing the problem on?

1.0-beta1

### Any related code or configuration details to produce this issue as well as relevant log output.

```shell

```

PR #89, opened 2025-03-22
88 feature add member management page

calebbourg:
## Description
This PR adds a `/organizations/{id}/members` page that will be used for CRUD operations on users.

As of right now the fetching logic is as follows:
- Fetch all users associated with the current organization
- Fetch all coaching relationships associated with the current organization
- filter coaching relationships by which ones are associated with the current user
- filter users by which ones are associated with the filtered list of coaching relationships

The goal is to only display users that have a direct relationship with the current user. 
This may change in the future, depending on how we want to manage users.

A few other changes of note:
- I moved the core data shape for a `User` into its own type and provided a "decorated" version for `UserSession`. Now that we are offering management of users, I think it make sense to allow the two to diverge as the need. I'm not 100% yet whether a `UserSession` should be a decorated version of `User` or if that relationship should be inverted, but this is a start.
- I modified `EntityApi.useEntityMutation` to be able to distinguish between the shape of data sent to our BE vs the shape of data returned. This made sense to me as we do not (and probably should not) send an `id` when creating a new entity as the `id` is something generated in the database by design. The returned data on creation _will_ have an `id`, however and so the types describing these two concepts should be allowed to be different where necessary. (For existing APIs I just provided the same type twice for backward compatibility Ex. `EntityApi.useEntityMutation<Action, Action>`.
- I started a pattern for nesting our API modules in the same way that the BE nests. Ex. users scoped to an organization. As it turns out, `createNested` was perfect for this. I explain a bit more about my thought process for this in https://github.com/refactor-group/refactor-platform-rs/pull/111

### Changes
* Adds member page
* Adds User API(s)

Screenshot 2025-03-23 at 2 33 19 PM
Screenshot 2025-03-23 at 2 33 14 PM

### Open Questions
- What is the best way to display users? A given user can be a Coach _and_ a Coachee. Do we want to display that type of information here and if so which role do we use?

- Partially addresses issue #80, provides a pattern to use for other entity types but doesn't apply it to every one.


src/components/user-auth-form.tsx
```diff
@@ -40,7 +40,7 @@ export function UserAuthForm({ className, ...props }: UserAuthFormProps) {
     };
 
     await createUserSession(userSession)
-      .then((userSession) => {
+      .then((userSession: UserSession) => {
```
calebbourg, line 43: This came up in typechecking

src/lib/api/actions.ts
```diff
@@ -169,7 +169,7 @@ export const useActionBySession = (coachingSessionId: Id) => {
  * error: Error object if the last operation failed, null otherwise
  */
 export const useActionMutation = () => {
-  return EntityApi.useEntityMutation<Action>(ACTIONS_BASEURL, {
+  return EntityApi.useEntityMutation<Action, Action>(ACTIONS_BASEURL, {
```
calebbourg, line 172: Here's the example where types `T` and `U` can be the same.

src/lib/api/entity-api.ts
```diff
@@ -362,15 +362,16 @@ export namespace EntityApi {
...
    * @returns Object with CRUD methods, loading state, and error state
    */
-  export const useEntityMutation = <T>(
+  export const useEntityMutation = <T, U>(
     baseUrl: string,
     api: {
-      create: (entity: T) => Promise<T>;
-      createNested: (id: Id, entity: T) => Promise<T>;
+      create: (entity: T) => Promise<U>;
+      createNested: (id: Id, entity: T) => Promise<U>;
```
calebbourg, line 374: This is where I changed this to allow us to provide differing types for what is sent vs what is received

src/types/user-session.ts
```diff
@@ -1,44 +1,22 @@
...
-    return false;
-  }
-  const object = value as Record<string, unknown>;
+import { User } from "@/types/user";
+/**
+ * This is an intersection type that combines the User type with additional properties.
+ * Currently, it does not add any new properties, but it can be extended in the future
+ * to decorate the User type if the two types need to diverge in the future.
+ */
+export type UserSession = User & {};
```
calebbourg, line 7: This is how I chose to have `UserSession` "wrap" `User` with the option of providing other fields if/when necessary. 
I asked cursor how best to do something like this in typescript and this is the result. Not sure if this is super idiomatic but it seems common in the ecosystem.

src/components/ui/members/add-member-dialog.tsx
```diff
@@ -0,0 +1,126 @@
...
+    e.preventDefault();
+
+    const newUser: NewUser = {
+      first_name: formData.firstName,
+      last_name: formData.lastName,
+      display_name: formData.displayName,
+      email: formData.email,
+    };
+
+    await createUserNested(currentOrganizationId, newUser);
```
jhodapp, line 57: You may want to import and consider calling `refresh()` after adding this, you'll get any other currently rendered list of Users automatically re-reading from the cache then.

Take a look at `addAction()` in actions-list.tsx for an example.

src/lib/api/entity-api.ts
jhodapp: I like this change very much, great addition!

src/lib/api/users.ts
```diff
@@ -1 +1,94 @@
...
+/**
+ * Hook for user mutations.
+ * Provides methods to create, update, and delete users.
+ */
+export const useUserMutation = () => {
+  return EntityApi.useEntityMutation<User, User>(USERS_BASEURL, {
+    create: UserApi.create,
+    createNested: async (id: Id, entity: User) => {
+      throw new Error("Create nested operation not implemented");
+    },
```
jhodapp, line 90: Minor thing, but do you mind defining createdNested above and putting the not-implemented exception in that function to be consistent with the other type APIs?

src/types/user-session.ts
jhodapp: I'm curious if the intersection with `{}` is necessary? According to [Perplexity's explanation](https://www.perplexity.ai/search/can-you-explain-this-typescrip-2MYQmj2URQWkygnMl8ZGyg), this might not be needed. I've not used Typescript's type intersection before though, so maybe it is necessary.

src/types/user.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export interface NewUser {
+  email: string;
+  password?: string;
+  first_name: string;
+  last_name: string;
+  display_name: string;
+}
+
+export type UserCategory = "coach" | "coachee" | "generalUser";
```
jhodapp, line 21: Nice, I like this.

jhodapp: Is "generalUser" redundant perhaps since it's a User type anyway? Can we just use "general" instead?

jhodapp (changes requested):
Looks great Caleb, I haven't been able to run this yet but I did a thorough code review of it. Let me know if you have any questions about my suggested changes.

src/types/user-session.ts
calebbourg: Yeah thank you for keeping me honest on this one:

Perplexity correctly identified my intention here:

> Why Use This Code?
> At first glance, this code might seem redundant because UserSession is effectively just an alias for User. > However, there are a few possible reasons why this pattern might be used:
>
> Future Extensibility:
>
>The developer might plan to extend the UserSession type later by adding more properties to it. For example:

With that in mind I think I'd prefer to leave it as is. The empty `{}`, although not necessary at the moment, provides a sure signal that "this is how we will extend this if needed in the future"

src/types/user.ts
calebbourg: Good question. I actually removed this in 74a470fd25c121bdcaf12677819f4f2aa363f67b as it was a hold over from a previous implementation for this that I tried. 
I could see it coming back at some point though at which time I think you're right that the "User" part is redundant.

src/lib/api/organizations/users.ts
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Hook for fetching a list of users.
+ */
+export const useUserList = (organizationId: Id) => {
+  return EntityApi.useEntityList<User>(
+    usersBaseUrl(organizationId),
+    () => UserApi.list(organizationId),
+    { organizationId }
+  );
```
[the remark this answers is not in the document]
calebbourg: Excellent suggestion 74a470fd25c121bdcaf12677819f4f2aa363f67b

src/lib/api/organizations/users.ts
```diff
@@ -0,0 +1,84 @@
+// Interacts with the organizations/{organizationId}/users endpoints
+
+import { Id } from "@/types/general";
+import { EntityApi } from "../entity-api";
+import { User, NewUser } from "@/types/user";
+import { ORGANIZATIONS_BASEURL } from "../organizations";
+
+const usersBaseUrl = (organizationId: Id) =>
```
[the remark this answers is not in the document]
calebbourg: Updated! 74a470fd25c121bdcaf12677819f4f2aa363f67b

src/app/organizations/[id]/members/page.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { MemberList } from "@/components/ui/members/member-list";
+import { useAuthStore } from "@/lib/providers/auth-store-provider";
+import { useCoachingRelationshipList } from "@/lib/api/coaching-relationships";
+import { useUserList } from "@/lib/api/organizations/users";
+import { useOrganizationStateStore } from "@/lib/providers/organization-state-store-provider";
+
+export default function MembersPage({
+  params,
+}: {
+  params: Promise<{ id: string }>;
```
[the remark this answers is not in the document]
Comment (calebbourg) [reply src/app/organizations/[id]/members/page.tsx:14]: Yup! 74a470fd25c121bdcaf12677819f4f2aa363f67b

src/components/ui/members/add-member-button.tsx
```diff
@@ -0,0 +1,23 @@
+"use client";
+
+import { Button } from "@/components/ui/button";
+import { Dialog, DialogTrigger } from "@/components/ui/dialog";
+import { AddMemberDialog } from "./add-member-dialog";
+import { useState } from "react";
+import { Plus } from "lucide-react";
+
+export function AddMemberButton() {
```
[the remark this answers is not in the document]
calebbourg: Makes way more sense. Thank you. I was very unsure in my naming 9aa56d99bd2ca2e943c1f28974a41c4061ded701

calebbourg:
@jhodapp I updated the component structure a bit to centralize the SWR cache management inside of `member-container` based on your suggestion to use `refresh()`
23fe24480f19426727e1c0e68a099519be25eef3

I would love for yours and @zgavin1 's feedback on this approach or how it can be improved. Still very much learning.

src/lib/api/users.ts
calebbourg: oh oops. Yup e33d4cfd6f4db903db77cd62af0b7526ae627c26

src/types/user-session.ts
jhodapp: Ok, I see you removed it in your latest commits, but also I see a comment which I think holds your intention of future expansion.

src/components/ui/members/add-member-button.tsx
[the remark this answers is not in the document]
jhodapp: You're welcome, I know it can be a little confusing sometimes. I've started to pick up on the patterns of naming from working with this over the last several months, so I'm happy to pass these learnings along.

jhodapp:
> @jhodapp I updated the component structure a bit to centralize the SWR cache management inside of `member-container` based on your suggestion to use `refresh()` [23fe244](https://github.com/refactor-group/refactor-platform-fe/commit/23fe24480f19426727e1c0e68a099519be25eef3)

I think it's a great approach, triggering the refetch of relationships and users list data in one go.

src/components/ui/members/member-container.tsx
```diff
@@ -0,0 +1,54 @@
+import { Card, CardContent } from "@/components/ui/card";
```
jhodapp, line 1: These seem to be unused in this file.

```suggestion
```

jhodapp (changes requested):
Just one minor tweak and I'm ready to approve.

src/components/ui/members/member-container.tsx
calebbourg: Good catch 1c1c86bfd8ab33f7f45819a79a06caaaf77cf48e

jhodapp (approved):
Looks great!

src/lib/api/entity-api.ts
zgavin1: Really like this  change - you could even have U default to T:

```
export const useEntityMutation = <T, U = T>(
```

because I see several places where they are still the same, but since we require two generics you have to assign two of the same types. Unless we envision that being more uncommon than it is now.

zgavin1 (approved):

PR #90, opened 2025-03-22
Add a collapsible sidebar, replacing the top menu-bar

### LINKED ISSUE #85: [Feature]: Add a sidebar to the layout
Add a sidebar component that contains:

1. A current organization logo/name drop down box at the top
2. Navigation buttons to Dashboard and Members pages
3. Eventually, Referrals, Organization settings and System status at the bottom

jhodapp:
## Description
This PR adds a new collapsible sidebar, replacing the top menu-bar. Incorporates setting the current Organization into the sidebar.

**Reviewing Note:** Adding the new components for the sidebar via the CLI also updated many of the other shadcn components, making the PR unexpectedly large. Review should focus mostly on the app-sidebar.tsx, organization-switcher.tsx, dashboard/layout.tsx, dashboard/page.tsx, coaching-sessions/[id]/layout.tsx, coaching-sessions/[id]/page.tsx

### Changes
* Adds a new collapsible sidebar component and current organization switcher
* Removes the mobile-nav menu since the sidebar has a mobile-friendly mode
* In organization-state-store.ts, we now set the `currentOrganizationId` when a complete `currentOrganization` is set

Dashboard

Jim Hodapp __ Caleb Bourg

### Testing Strategy
1. Log in
2. Observe the new sidebar on the dashboard and coaching session pages
3. You should be able to select your organization from the new organization-switcher in the sidebar
4. The sidebar should collapse into an icon-only version of itself when clicking the collapse button
5. It should be responsive to all screen sizes

### Concerns
1. Custom Organization logos aren't working yet


calebbourg (commented):
@jhodapp I tested this out locally and it looks great!! Nice job.
Without going too much into the code, I wanted to ask about the intend behavior when a user chooses a new organization from the Coaching Sessions page. 
Right now it does not appear to change anything. What would change in the future? It seems like it should do something like kick you back out to the dashboard with the new organization selected? 
Or maybe a user should not be able to choose a different organization directly from the Coaching Session page? 
Curious what your thoughts are

jhodapp:
That's a great question, I hadn't thought about that. Let me think on that a bit.

jhodapp:
@calebbourg I think it makes the most sense to optionally display/hide the OrganizationSwitcher component on the sidebar. It definitely makes sense to show it on the Dashboard page, but it doesn't make sense to show on the Coaching Sessions page, and selecting another organization and having it take you then to the Dashboard page seems a little unexpected.

Thoughts?

calebbourg:
@jhodapp I think that makes absolute sense. Another question just for the sake of thinking through it, is what value does it add to the sidebar on a page where you already can choose an organization (dashboard)? Just thinking the extra logic of hiding it and showing it at appropriate times might be avoided if we deem it redundant. 
That's my naive thoughts though. I don't have the full context of your vision with it.

calebbourg (approved):

PR #92, opened 2025-03-26
88 feature add member management page add users

calebbourg:
## Description
This PR adds two main things:
- Completes the `members` link between `/dashboard` and `/organizations/[id]/members`
- Completes the "Add new member" button

**Note:** Only users that are a coach to _some_ other user within the current organization will be able to create new users. The newly created user will always be created as a coachee to the current user at this time.

### Changes
* Allows coaches to create new coachees

### Testing Strategy
Tested locally against https://github.com/refactor-group/refactor-platform-rs/pull/113

### DIFF
--- a/src/components/ui/members/add-member-dialog.tsx
+++ b/src/components/ui/members/add-member-dialog.tsx
@@ -175,9 +175,11 @@ export function AddMemberDialog({
               <div className="text-red-500 text-sm">{passwordError}</div>
             )}
           </div>
-          <DialogFooter>
-            <Button type="submit">Create Member</Button>
-          </DialogFooter>
+          <div className="pt-4">
+            <DialogFooter>
+              <Button type="submit">Create Member</Button>
+            </DialogFooter>
+          </div>
         </form>
       </DialogContent>
     </Dialog>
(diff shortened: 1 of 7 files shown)


src/components/ui/members/add-member-dialog.tsx
```diff
@@ -52,80 +58,126 @@ export function AddMemberDialog({
...
+            </div>
+            {passwordError && (
+              <div className="text-red-500 text-sm">{passwordError}</div>
+            )}
           </div>
-          <Button type="submit" className="w-full">
-            Add Member
-          </Button>
+          <DialogFooter>
+            <Button type="submit">Create Member</Button>
```
zgavin1, line 179: I think it would be good UX here to disable form submission button until all input fields are non-empty at least.

jhodapp: Indeed, also it looks like the button needs some padding added to it to pad it from the confirm password textbox.

Screenshot 2025-03-28 at 16 23 37

jhodapp (changes requested):
Some minor UI tweaking needed, otherwise looking great.

calebbourg:
@jhodapp How does this look 255b89e8142b11cf6187a313359b877bc5750ca5

Screenshot 2025-03-29 at 8 56 14 AM

jhodapp (approved):

PR #96, opened 2025-04-02
Update coaching sessions

calebbourg:
## Description
This PR adds the ability to update an existing Coaching Session.

It also adds a new type of menu to the coaching session component. Open to suggestion on if this is what we want.

### Changes
* Updated the `add-new-coaching-session` dialog component to be re-usable for creating new coaching session and updating existing coaching sessions
* Adds a "3 dot" menu to the coaching sessions component
* Adds a "Delete Coaching Session" link that will be used later

Screenshot 2025-04-02 at 6 56 35 AM

### Testing Strategy
tested locally

### Concerns


src/components/ui/dashboard/coaching-session-dialog.tsx
```diff
@@ -0,0 +1,150 @@
...
+              value={sessionTime}
+              onChange={(e) => setSessionTime(e.target.value)}
+              className="w-full border rounded p-2"
+              required
+            />
+          </div>
+          <Button type="submit" disabled={!sessionDate || !sessionTime}>
+            {mode === "create" ? "Create Session" : "Update Session"}
+          </Button>
+        </form>
```
zgavin1, line 146: We probably want the form itself (and its submit logic) moved into it's own react component and then composed into the dialog component. Separates concerns of the dialog UI vs the form pretty nicely that way.

```
<CoachingSessionDialog ...props >
    <CoachingSessionForm ...props />
</CoachingSessionDialog>
```

src/lib/api/actions.ts
```diff
@@ -65,6 +65,17 @@ export const ActionApi = {
...
     EntityApi.deleteFn<null, Action>(`${ACTIONS_BASEURL}/${id}`),
+
+  /**
+   * Deletes an action nested under another entity (foreign key relationship).
+   *
+   * @param entityId The ID of the entity under which to delete the action
+   * @param actionId The ID of the action to delete
+   * @returns Promise resolving to the deleted Action object
+   */
+  deleteNested: async (_entityId: Id, _actionId: Id): Promise<Action> => {
```
zgavin1, line 76: forgive me I just haven't noticed this until now - what are the underscore prefixes on the params? Just acknowledging the api method isn't implemented yet?

src/lib/api/coaching-sessions.ts
```diff
@@ -95,6 +95,20 @@ export const CoachingSessionApi = {
...
+  /**
+   * Deletes a coaching session nested under another entity (foreign key relationship).
+   *
+   * @param entityId The ID of the entity under which to delete the coaching session
+   * @param coachingSessionId The ID of the coaching session to delete
+   * @returns Promise resolving to the deleted CoachingSession object
+   */
+  deleteNested: async (
+    _entityId: Id,
+    coachingSessionId: Id
```
zgavin1, line 108: does this need the underscore?

jhodapp:
@calebbourg Thanks for this, it's nice to see this functionality in place!!

Can you move the "Join Session" back out into a primary button that sits just to the left of the new vertical "..." menu and change the text to "Join"? And then change "Edit Session" and "Delete Session" to "Edit" and "Delete"?

src/components/ui/coaching-session.tsx
```diff
@@ -1,18 +1,30 @@
...
+import { DateTime } from "ts-luxon";
 
 interface CoachingSessionProps {
   coachingSession: {
     id: Id;
     date: string;
+    coaching_relationship_id: Id;
+    created_at: DateTime;
+    updated_at: DateTime;
   };
```
jhodapp, line 28: Change this to use the type defined in `src/types/coaching-session.ts`

src/components/ui/coaching-session.tsx
```diff
@@ -22,29 +34,58 @@ const CoachingSession: React.FC<CoachingSessionProps> = ({
...
-      </CardHeader>
-    </Card>
+        </CardHeader>
+      </Card>
+      <CoachingSessionDialog
+        open={updateDialogOpen}
+        onOpenChange={setUpdateDialogOpen}
+        onCoachingSessionUpdated={() => {
+          // Refresh the list of coaching sessions
+          window.location.reload();
```
jhodapp, line 83: Is there a hook for the coaching sessions that we can call `refresh()` on instead of this? I may be confused on what this is for otherwise.

src/components/ui/dashboard/coaching-session-dialog.tsx
jhodapp: I think that's a sensible suggestion Zach.

src/components/ui/dashboard/coaching-session-dialog.tsx
```diff
@@ -0,0 +1,150 @@
...
+          updated_at: DateTime.now(),
+        });
+      }
+      refresh();
+      onCoachingSessionUpdated();
+      setSessionDate(undefined);
+      setSessionTime("");
+      onOpenChange(false);
+    } catch (error) {
+      console.error(`Failed to ${mode} coaching session:`, error);
```
jhodapp, line 97: Would you mind adding a TODO comment here saying to also display an error as a toast element to the user once we get toast infrastructure in place?

jhodapp (changes requested):
Looking great, a few inline suggestions.

src/components/ui/dashboard/coaching-session-dialog.tsx
calebbourg: added! e0074c6281903cfd0496013ff251999b56c24be7

calebbourg:
@jhodapp I moved the join session button back out in e18bfed006f54aa958168ed8416530905dae9eda
Screenshot 2025-04-04 at 8 27 39 AM

Let me know how this works for now.

jhodapp:
> @jhodapp I moved the join session button back out in [e18bfed](https://github.com/refactor-group/refactor-platform-fe/commit/e18bfed006f54aa958168ed8416530905dae9eda) Screenshot 2025-04-04 at 8 27 39 AM
> 
> Let me know how this works for now.

@calebbourg 

Closer, but now I see the "Create New Coaching Session" button repeated above every session. Are you seeing this too?

Screenshot 2025-04-04 at 12 19 04

jhodapp:
@calebbourg Your latest changes also seem to degrade the AddEntity component up top on the Dashboard page as well:

Screenshot 2025-04-04 at 12 23 15

calebbourg:
@jhodapp oh haha for some reason I thought that had something to do with changes you were making on the dashboard being in an in between state.. I see now that that would be really weird :) I'll fix it

calebbourg:
@jhodapp @zgavin1 I took another crack at this. I realized that we were using the `CoachingSessionDialog` component in multiple places throughout the larger dashboard page and so attempted to declare it once and provide hooks for the various other components to call into it's functionality. I defined the hooks in the top level dashboard page. In order to do that I needed to separate some of the code into a client rendered function. that is called by `page.tsx` which remains server-rendered. I _think_ that's a reasonable approach but open to thoughts.

Another spin-off of this was that I removed the "Create Coaching Session" button from the actual list of coaching sessions as It seemed redundant given the `add-entities` buttons above.

jhodapp:
> Another spin-off of this was that I removed the "Create Coaching Session" button from the actual list of coaching sessions as It seemed redundant given the `add-entities` buttons above.

@calebbourg this is awesome, it was something I was going to do anyway when I moved the coaching sessions list down in the page for exactly the reason you mentioned. Thanks for doing this!

src/app/dashboard/dashboard-content.tsx
```diff
@@ -0,0 +1,69 @@
...
+  };
+
+  return (
+    <>
+      <div className="p-4 max-w-screen-2xl">
+        <div className="mb-8 w-full">
+          <AddEntities onCreateSession={() => handleOpenDialog()} />
+        </div>
+      </div>
+      <DashboardContainer>
```
jhodapp, line 57: I'd like to suggest that we should merge `DashboardContainer` and `DashboardContent` and I suggest `DashboardContainer` as the name since we have precedence for this already. Unless I'm misunderstanding things, I don't see any value or requirement to have the extra div layer.

I do think you did take the right approach though in making the Dashboard page as simple as possible and trying to keep it rendered on the server side.

src/components/ui/dashboard/add-entities.tsx
```diff
@@ -28,27 +25,16 @@ export default function AddEntities() {
...
-          onOpenChange={setOpen}
-          onCoachingSessionUpdated={onCoachingSessionAdded}
-          dialogTrigger={
-            <AddCoachingSessionButton
-              disabled={!isCoach || !currentOrganizationId}
-            />
-          }
+        <AddCoachingSessionButton
+          disabled={!isCoach || !currentOrganizationId}
+          onClick={onCreateSession}
```
jhodapp, line 30: Yes!! This fulfills my TODO item, so thank you for refactoring this to match the style of AddMemberButton. What a pleasant surprise to discover. :)

src/components/ui/coaching-session.tsx
```diff
@@ -15,56 +15,49 @@ import {
...
-                    href={`/coaching-sessions/${coachingSession.id}`}
-                    onClick={() =>
-                      setCurrentCoachingSessionId(coachingSession.id)
-                    }
-                  >
-                    Join Session
-                  </Link>
-                </DropdownMenuItem>
-                <DropdownMenuItem onClick={() => setUpdateDialogOpen(true)}>
+                <DropdownMenuItem onClick={onUpdate}>
```
jhodapp, line 60: To be consistent with Agreements and Actions, can you update the drop down menu item text to be "Edit" and "Delete" instead of "Update Session" and "Delete Session"?

jhodapp (changes requested):
Such an awesome change set, thank you Caleb. Just a few more suggestions and I'll be ready to approve.

src/app/dashboard/dashboard-content.tsx
jhodapp: I did some updating here, but not all of what I suggested. But you'll want to check it out so we don't step on each other directly: https://github.com/refactor-group/refactor-platform-fe/pull/102/files#diff-1b7538b3757a933d976e1a796af3bbf3bc3895ae83f60d5a0562db4d2f6bbc02

src/components/ui/coaching-session.tsx
calebbourg: Updated 536a763ba3acddbbfb02dc1dfae19c24dc55447b

src/app/dashboard/dashboard-content.tsx
calebbourg: ok made this change e3ef1f5ffb59a839de82b756046e15336aa7e083

jhodapp (approved):
Awesome set of changes, thanks Caleb!

PR #97, opened 2025-04-02
Leverage default generic useEntityMutation function definition

zgavin1:
## Description
Implement a default generic type for the second argument of the `useEntityMutation` hook, makes it cleaner to use across the different entity specific-hooks which may or may not need the second type.

### Changes
* implement the type definition change
* update usages of this hook that were providing the extra duplicate arg.

### Testing Strategy

### Concerns
Not totally sure why this didn't work when Caleb tried - may have just been a typechecker hiccup that day.


calebbourg (approved):
@zgavin1 Nice job! I must have done something weird when I was trying this!

PR #103, opened 2025-04-08
Coaching session page tweaks

jhodapp:
## Description
This PR tweaks the coaching session page layout padding and makes the CoachingSessionSelector use React API hooks instead of synchronous fetch calls.

### Changes
* Add some padding to the coaching session page title left side and from the main-nav header
* Restructure the CoachingSessionSelector to use the API hooks instead of useEffect blocks, which allows the list to be updated with a call to refresh() from the OverarchingGoalContainer

Screenshot 2025-04-07 at 21 16 02

### Testing Strategy
1. Observe the left padding shift to the right to the session title on the coaching session page
2. Update a coaching session overarching goal, observe that the coaching session selector now updates the overarching goal in the list and the selected one once you click on the drop down

### Concerns
* I still want to get it so that the coaching session selector will update without having to click on it, but I can't quite figure that out yet. This is a step in the right direction however.

### DIFF
--- a/src/components/ui/coaching-session-selector.tsx
+++ b/src/components/ui/coaching-session-selector.tsx
@@ -12,19 +12,18 @@ import {
 } from "@/components/ui/select";
 import { getDateTimeFromString, Id } from "@/types/general";
 import { useCoachingSessionList } from "@/lib/api/coaching-sessions";
-import { useEffect, useState } from "react";
+import { useEffect } from "react";
 import { DateTime } from "ts-luxon";
 import { useCoachingSessionStateStore } from "@/lib/providers/coaching-session-state-store-provider";
-import { OverarchingGoalApi } from "@/lib/api/overarching-goals";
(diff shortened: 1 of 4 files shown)


src/components/ui/coaching-session-selector.tsx
```diff
@@ -145,6 +84,33 @@ function CoachingSessionsSelectItems({
...
+  return (
+    <SelectItem value={session.id}>
+      <div className="flex min-w-0">
+        <div className="min-w-0 w-full">
+          <p className="truncate text-sm font-medium">
+            {overarchingGoal.title || "No goal set"}
+          </p>
+          <p className="truncate text-sm text-gray-400">
+            {getDateTimeFromString(session.date).toLocaleString(
+              DateTime.DATETIME_FULL
```
calebbourg, line 105: Does this use the local machine's timezone?

calebbourg (approved):

PR #109, opened 2025-04-13
Fix issue with link that is also strikethrough, and fix outdated link tooltip text

zgavin1:
## Description
Tooltip text was wrong

Also when text is both a link and has strikethrough styling, shift left click was not working

Screenshot 2025-04-12 at 7 47 27 PM


jhodapp (approved):
Works perfectly, thanks Zach!

PR #132, opened 2025-06-18
Handle an invalid login error case

### LINKED ISSUE #122: [Bug]: Unsuccessful login returns 500
### Contact Details

james.hodapp@gmail.com

### What happened?

Providing incorrect login credentials returns a 500 error instead of a more friendly "Invalid username or password" error.

### What is the version of refactor-platform-fe that you're seeing the problem on?

1.0.0-beta1

### Any related code or configuration details to produce this issue as well as relevant log output.

```shell

```

jhodapp:
## Description
This PR fixes the issue where the login form wasn't displaying the correct error when passing an invalid credential. 

### Changes
* Moves `user-auth-form.tsx` under `components/ui/login` to be consistent with other component src tree layout
* Also adds a new error type EntityApiError that extends Error, because we weren't handling error conditions while using EntityApi otherwise. This was needed to successfully handle the invalid login error case 

Screenshot 2025-06-17 at 20 31 13

### Testing Strategy
1. You need the [backend PR here first](https://github.com/refactor-group/refactor-platform-rs/pull/160)
2. Then, try logging in with any random credentials. You should see the "Invalid email or password. Please try again." error displayed in red on the login page.

### Concerns
None

### DIFF
--- a/src/types/general.ts
+++ b/src/types/general.ts
@@ -1,7 +1,101 @@
 import { DateTime } from "ts-luxon";
+import { AxiosError, AxiosResponse } from "axios";
+import axios from "axios";
 
 // A type alias for each entity's Id field
 export type Id = string;
+
+/**
+ * Enhanced error type for Entity API operations that preserves axios error information
+ * while providing a more user-friendly interface for error handling.
+ */
+export class EntityApiError extends Error {
+  public readonly method: string;
+  public readonly url: string;
+  public readonly status?: number;
+  public readonly statusText?: string;
+  public readonly response?: AxiosResponse;
+  public readonly request?: any;
+  public readonly data?: any;
+  public readonly originalError: AxiosError | Error;
+
+  constructor(
+    method: string,
+    url: string,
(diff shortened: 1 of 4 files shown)


src/types/general.ts
```diff
@@ -1,8 +1,102 @@
...
+   */
+  public isServerError(): boolean {
+    return this.status !== undefined && this.status >= 500;
+  }
+
+  /**
+   * Returns true if this error represents a network error (no response received)
+   */
+  public isNetworkError(): boolean {
+    return this.status === undefined && "isAxiosError" in this.originalError;
```
calebbourg, line 96: Curious how this works?

calebbourg (approved):

src/types/general.ts
jhodapp: @calebbourg It's basically the "other" kind of error being labeled as a network error. If we see opportunity to tighten this up with specific errors in the future, we can definitely do so.

PR #151, opened 2025-07-05
150 introduce a testing framework and some high impact initial tests

### LINKED ISSUE #150: Introduce a testing framework and some high-impact initial tests
We have come to the point where we need to introduce a testing framework and some high-impact initial tests for the frontend.

Why now?

1. Manual testing is getting too complex and time consuming to do.
2. Some of the logic, for example having correct set/update behavior of `currentOrganizationId`, `currentCoachingRelationshipId`, and `currentCoachingSessionId` across pages, across multiple browsers tabs has grown complex enough, yet very important, to ensure correctness.
3. It'll become too hard to introduce if we wait any longer, even with the help of AI coding.

jhodapp:
## Description
This PR introduces [Vitest](https://vitest.dev), [React Testing library](https://testing-library.com/docs/react-testing-library/intro), [Playwright](https://playwright.dev), and [MSW](https://mswjs.io) for unit, integration, and e2e testing along with initial tests to ensure state store logic around the stored entity Ids functions correctly (i.e. current organization, current coaching relationship, current coaching session).

### Changes
* Add vitest for unit testing of functions, hooks, and components
* Add react testing library for integration testing
* Add playwright for cross-browser (desktop and mobile) e2e user flow testing
* Add msw for mocking backend API calls
* Updates our node.js.yml Build & Test GitHub Action workflow to run the tests
* Update to latest stable next.js and react versions. Turbopack is no longer experimental, so use it as stable.

### Testing Strategy
1. Run `npm run test:all:run` ensuring that all tests pass

### Concerns
None


.github/workflows/node.js.yml
```diff
@@ -11,22 +14,66 @@ on:
...
       uses: actions/setup-node@v3
       with:
         node-version: ${{ matrix.node-version }}
         cache: 'npm'
-    - run: npm ci
-    - run: npm run build --if-present
-    # TODO: once we have a test runner/framework in place, uncomment this
-    # - run: npm test
+
+    - name: Download build artifacts
```
calebbourg, line 64: Where is this downloaded to and what are we using it for? (or my own education)

src/components/ui/organization-switcher.tsx
```diff
@@ -70,12 +70,10 @@ export function OrganizationSwitcher({
       organizations &&
       organizations.length > 0
     ) {
-      console.trace(
```
calebbourg, line 73: What was the reason for removing these? Were they for debugging only?

src/test-utils/msw-handlers.ts
```diff
@@ -0,0 +1,49 @@
...
+        id: params.id,
+        title: 'Session #1',
+        coaching_relationship_id: 'rel-1',
+        scheduled_date: '2025-07-04T10:00:00Z',
+      }
+    })
+  }),
+
+  // Mock coaching relationships API
+  http.get('/api/organizations/:orgId/coaching_relationships', () => {
```
calebbourg, line 27:
```suggestion
  http.get('/api/organizations/:organizationId/coaching_relationships', () => {
```

I think? For consistency

src/test-utils/providers.tsx
```diff
@@ -0,0 +1,16 @@
+import { ReactNode } from 'react'
+import { AuthStoreProvider } from '@/lib/providers/auth-store-provider'
+import { OrganizationStateStoreProvider } from '@/lib/providers/organization-state-store-provider'
+import { CoachingRelationshipStateStoreProvider } from '@/lib/providers/coaching-relationship-state-store-provider'
+
+export function TestProviders({ children }: { children: ReactNode }) {
```
calebbourg, line 6: It wasn't obvious until I saw this file what `TestProviders` was accomplishing. Now it makes total sense. 
I wonder if it would make sense to just provide an application level aggregation of these that can be used in the app _and_ in the tests? Like a `Providers`?

calebbourg (commented):
This is fantastic! Thank you. Just left some questions

.github/workflows/node.js.yml
```diff
@@ -11,22 +14,66 @@ on:
 
 jobs:
   build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [21.x]
```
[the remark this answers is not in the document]
jhodapp: Agreed, simplifying to pin it to a non-matrix strategy.

.github/workflows/node.js.yml
jhodapp: That's no longer part of the build process since we're not trying to cache anything in between a multi-version matrix-strategy build process.

src/test-utils/providers.tsx
jhodapp: A really sharp observation and definitely worth doing.

PR #204, opened 2025-10-15
Update to user UserRole

calebbourg:
## Description
  Implement organization-scoped user roles to support multi-organization access and SuperAdmin privileges. Users can now have different roles in different organizations, with SuperAdmin having global access across all organizations.

  #### GitHub Issue: Resolves #185

  ### Changes
  * Added `UserRole` interface and `roles: UserRole[]` field to `User` type to support organization-scoped role assignments
  * Added `SuperAdmin` role to `Role` enum for global system access
  * Created `getUserRoleForOrganization()` utility function to determine user's role based on organization context
  * Created `useCurrentUserRole()` hook to provide context-aware role resolution
  * Updated `member-container.tsx` and `member-card.tsx` to use organization-scoped roles instead of deprecated `user.role`
  * Added `.serena/` to `.gitignore` to exclude serena configuration files

  ### Testing Strategy
Regression Tested Locally

  ### Concerns
  * The legacy `user.role` field is now deprecated but still present in the User type - will need backend migration to fully remove it
  * Error handling throws when no matching role found - may want to add graceful fallback or redirect to organization selection

### DIFF
--- a/__tests__/components/ui/tiptap-ui/link-popover.test.tsx
+++ b/__tests__/components/ui/tiptap-ui/link-popover.test.tsx
@@ -1,0 +1,448 @@
+import { render, screen, waitFor } from '@testing-library/react'
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import userEvent from '@testing-library/user-event'
+import { EditorProvider } from '@tiptap/react'
+import StarterKit from '@tiptap/starter-kit'
+import Link from '@tiptap/extension-link'
+import { LinkPopover } from '@/components/ui/tiptap-ui/link-popover'
+import { TooltipProvider } from '@/components/ui/tooltip'
+
+// Mock ResizeObserver for test environment
+global.ResizeObserver = vi.fn().mockImplementation(() => ({
+  observe: vi.fn(),
+  unobserve: vi.fn(),
+  disconnect: vi.fn(),
+}))
+
+describe('LinkPopover', () => {
+  const createEditorWithContent = (content: string = '') => {
+    const extensions = [
+      StarterKit,
+      Link.configure({
+        openOnClick: false,
+        HTMLAttributes: {
+          class: 'text-blue-600 underline',
+        },
+      }),
+    ]
+
+    return { extensions, content }
+  }
+
+  // Helper to wrap components with required providers
+  const renderWithProviders = (ui: React.ReactElement) => {
+    return render(
+      <TooltipProvider>
+        {ui}
+      </TooltipProvider>
+    )
+  }
+
+  beforeEach(() => {
+    // Clear any existing popovers from previous tests
+    document.body.innerHTML = ''
+  })
+
+  describe('Single Popover Rendering', () => {
+    it('should render only one popover when link button is clicked', async () => {
+      const user = userEvent.setup()
+      const { extensions, content } = createEditorWithContent('Test content')
+
+      renderWithProviders(
+        <EditorProvider extensions={extensions} content={content}>
+          <div data-testid="toolbar">
+            <LinkPopover />
+          </div>
+        </EditorProvider>
+      )
+
+      // Click the link button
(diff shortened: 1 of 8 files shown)


src/lib/hooks/use-current-user-role.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Determines the appropriate role based on:
+ * - SuperAdmin users (organization_id === null) have global access
+ * - Otherwise, finds the role matching the current organization ID
+ *
+ * @returns The user's Role for the current organization
+ * @throws Error if no matching role is found for the organization
+ */
+export const useCurrentUserRole = (): Role => {
+  const { userSession } = useAuthStore((state) => ({ userSession: state.userSession }));
+  const { currentOrganizationId } = useCurrentOrganization();
```
jhodapp: **Issue:** currentOrganizationId can be null or undefined (from useCurrentOrganization), but it's passed directly to `getUserRoleForOrganization` which expects `Id | null`. Need to ensure consistent null handling.

**Suggested Fix:** Add explicit null handling:
```typescript
export const useCurrentUserRole = (): Role | null => {
  const { userSession } = useAuthStore((state) => ({ userSession: state.userSession }));
  const { currentOrganizationId } = useCurrentOrganization();

  return useMemo(() => {
    if (!userSession?.roles || userSession.roles.length === 0) {
      return null;
    }

    // Ensure we have a valid organization context
    if (currentOrganizationId === undefined) {
      console.warn('No organization context available');
      return null;
    }

    try {
      return getUserRoleForOrganization(userSession.roles, currentOrganizationId);
    } catch (error) {
      return null;
    }
  }, [userSession?.roles, currentOrganizationId]);
};
```typescript

File: src/lib/hooks/use-current-user-role.ts [hunk 2]
```diff
@@ -0,0 +1,29 @@
+"use client";
+
+import { useMemo } from 'react';
+import { useAuthStore } from '@/lib/providers/auth-store-provider';
+import { useCurrentOrganization } from './use-current-organization';
+import { getUserRoleForOrganization, Role } from '@/types/user';
+
+/**
```
jhodapp: **Issue:** Hook doesn't document behavior when user switches organizations or when organization data is loading.

**Suggested Fix:** Add comprehensive JSDoc:
```typescript
/**
 * Hook that provides the user's role for the current organization context.
 *
 * @returns The user's Role for the current organization, or null if:
 *   - User has no roles assigned
 *   - No organization context is available
 *   - User doesn't have access to the current organization
 *
 * @example
 * const role = useCurrentUserRole();
 * if (role === Role.Admin) {
 *   // Show admin UI
 * }
 */
```

src/lib/hooks/use-current-user-role.ts
```diff
@@ -0,0 +1,29 @@
...
+  const { userSession } = useAuthStore((state) => ({ userSession: state.userSession }));
+  const { currentOrganizationId } = useCurrentOrganization();
+
+  return useMemo(() => {
+    if (!userSession?.roles || userSession.roles.length === 0) {
+      throw new Error('User has no roles assigned');
+    }
+
+    return getUserRoleForOrganization(userSession.roles, currentOrganizationId);
+  }, [userSession?.roles, currentOrganizationId]);
```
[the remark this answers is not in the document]
calebbourg: Added toast notifications in 9bb9451f31dd9196b4c428ab556d8053b0dd1890

src/lib/hooks/use-current-user-role.ts
calebbourg: Updated to reflect the new `UserRoleState` object as well 8fcdec260beb3a597567f192310e043899cfcbc7

calebbourg:
@jhodapp I believe I had addressed all of your comments. Thank you very much!

jhodapp (approved):

PR #212, opened 2025-10-17
User Role UI Updates

calebbourg:
## Description
  Enhance the members page with improved user identification, role visibility, and organizational context. The current user is
  now listed first with a "(You)" label, roles are displayed in alphabetical order combining organization and coaching
  relationship roles, and each member shows their coaches within the organization.

  #### GitHub Issue: Closes #185

  ### Changes
  * Sort members list to display current user first
  * Add "(You)" label next to current user's name
  * Display combined organization and coaching relationship roles for each member (alphabetically sorted)
  * Show coaches for each member under their role information
  * Add organization name subtext under "Members" heading ("of {Organization Name}")
  * Create reusable utility functions in `src/lib/utils/user-roles.ts`:
    - `getUserDisplayRoles()` - Combines organization roles and coaching roles
    - `getUserCoaches()` - Extracts coach names for a user
  * Add comprehensive unit tests (13 tests, all passing) covering all utility functions and edge cases
  * Fix SuperAdmin role display (handles `organization_id: null` correctly)

Screenshot 2025-10-17 at 7 08 12 AM

  ### Testing Strategy
  - All changes tested locally and verified working
  - 13 unit tests added for `user-roles.ts` utilities, all passing
  - Manual testing performed:
    - Verified current user appears first in members list
    - Confirmed "(You)" label displays correctly
    - Tested role combinations (Admin, User, SuperAdmin, Coach, Coachee)
    - Verified alphabetical sorting of roles
    - Confirmed coaches display correctly for users with 0, 1, or multiple coaches
    - Tested SuperAdmin role display with `organization_id: null`
    - Verified organization name appears in subtext

  ### Concerns
  None


src/lib/utils/user-roles.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  // Add coaching roles
+  const isCoach = relationships.some(r => r.coach_id === user.id);
+  const isCoachee = relationships.some(r => r.coachee_id === user.id);
+
+  if (isCoach) roles.add('Coach');
+  if (isCoachee) roles.add('Coachee');
+
+  return Array.from(roles).sort();
```
jhodapp: Or maybe we need a new `roles` types under `src/types/` that the `users` type can use? You could add a sort method in there.

src/components/ui/members/member-list.tsx
jhodapp: Ever though this isn't technically something you changed in this PR, I still wanted to pass along Claude's Typescript-focused PR review suggestion:

**Issue: Inconsistent ID Type Usage**

```typescript
const userRelationshipsMap = users.reduce((accumulator_map, user) => {
  const userRelationships = relationships.filter(
    (relationship) =>
      relationship.coach_id === user.id || relationship.coachee_id === user.id
  );
  accumulator_map[user.id] = userRelationships;  // ⚠️ Index signature issue
  return accumulator_map;
}, {} as { [key: string]: CoachingRelationshipWithUserNames[] });
```

**Problem:** Using Id (which is string) as index is fine, but the explicit type annotation { [key: string]: ...} is redundant.

Better approach:
```typescript
const userRelationshipsMap = users.reduce<Record<Id, CoachingRelationshipWithUserNames[]>>(
  (accumulator_map, user) => {
    const userRelationships = relationships.filter(
      (relationship) =>
        relationship.coach_id === user.id || relationship.coachee_id === user.id
    );
    accumulator_map[user.id] = userRelationships;
    return accumulator_map;
  },
  {}
);
```

**Benefits:**
- More explicit about using Id type
- Matches the pattern throughout the codebase
- Clearer intent

src/lib/utils/user-roles.ts
```diff
@@ -0,0 +1,54 @@
+import { User } from "@/types/user";
+import { CoachingRelationshipWithUserNames } from "@/types/coaching_relationship";
+import { Id } from "@/types/general";
+
```
jhodapp: **Readonly for Display Data**

Since display roles and coaches are derived data (not mutated), mark as readonly:

```typescript
export function getUserDisplayRoles(
  user: User,
  organizationId: Id,
  relationships: CoachingRelationshipWithUserNames[]
): readonly string[] {
  // ... implementation
  return Object.freeze(Array.from(roles).sort());
}

export function getUserCoaches(
  userId: Id,
  relationships: CoachingRelationshipWithUserNames[]
): readonly string[] {
  return Object.freeze(
    relationships
      .filter(r => r.coachee_id === userId)
      .map(r => `${r.coach_first_name} ${r.coach_last_name}`)
  );
}
```

**Benefits:**
- Prevents accidental mutation
- Signals immutability at type level
- Better performance hints to runtime

jhodapp (changes requested):
Very cool work here Caleb. Thank you brother!

**Here are my thoughts and recommendations for changes:**
1. Alphabetize the list of members while still keeping the "you" user as the first one
2. In a scenario where I have two organizations that I'm a member of while the rest of the users are not a member of the second one, only the first, the list of members is not just my user alone. It shows some other users. We suspected in our call today that the query that retrieves this list may not be filtering properly, perhaps only taking coaching relationship into account without taking `organization_users` correctly into account.

**Refactor Group:**

Image

**MotoAcquire:**

Image

_Note that I see three of those toast messages "You don't have access to this organization" even though I am a member of this organization in the `organization_users` table_

4. When a user does not currently have a coach assigned, display "None" for their list of coaches instead of hiding the "Coaches: " line entirely.
5. When a user is not an admin and they can't take any action in the Members page, let's not render the "..." menus since they're empty anyway. E.g.:
Image

5. When I tried assigning your user (Caleb) as the coach of Test User, I got a toast message saying that it successfully assigned you as their coach, but I got a Next.js HTTP 403 error saying I don't have permission to assign this even though I'm an organization admin.

Image

src/lib/utils/user-roles.ts
```diff
@@ -0,0 +1,54 @@
...
+  // Add organization-specific role
+  const orgRole = user.roles.find(r => r.organization_id === organizationId);
+  if (orgRole) {
+    roles.add(orgRole.role);
+  }
+
+  // Add coaching roles
+  const isCoach = relationships.some(r => r.coach_id === user.id);
+  const isCoachee = relationships.some(r => r.coachee_id === user.id);
```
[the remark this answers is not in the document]
calebbourg: great suggestion! [9285529](https://github.com/refactor-group/refactor-platform-fe/pull/212/commits/92855297086f49ac62c8b7cf0211c2061906d8a2)

jhodapp (approved):
Looks great and I think we got all of the major edge-cases taken care of. I know this one wasn't easy, so, thank you for your extra effort and attention to detail.

PR #238, opened 2025-12-09
SSE Frontend Implementation

calebbourg:
## Description

  Implements Server-Sent Events (SSE) infrastructure for real-time updates across coaching sessions. Users now see changes automatically
  when their coach or coachee creates/updates actions, agreements, or overarching goals - no manual refresh required.

  The implementation leverages SWR cache invalidation for automatic UI updates, requires zero changes to existing components, and follows
  established codebase patterns (Zustand, SWR, hooks, providers).

  ### Changes

  **SSE Infrastructure:**
  * Add SSE event type definitions with discriminated unions matching backend serialization
  * Create SSE connection state management with Zustand
  * Implement native EventSource connection with automatic reconnection
  * Add type-safe event handler hook with date transformation
  * Integrate SSE cleanup with logout registry

  **Real-time Cache Invalidation:**
  * Implement SWR cache invalidation for all SSE events (actions, agreements, goals)
  * Handle both string and array SWR cache keys
  * Add force logout system event handler

  **Provider Integration:**
  * Add SSE provider to root application providers
  * Use idiomatic Zustand pattern with useState for store initialization
  * Fix unstable useEffect dependencies causing connection instability

  **Security & Code Quality:**
  * Remove sensitive data logging (user IDs, session IDs, content) from console
  * Add comprehensive SSE implementation documentation
  * Standardize OverarchingGoal date fields to DateTime

  ### Testing Strategy

  **Manual Testing (Two Browser Sessions):**
  1. Log in as User A in one browser, User B in another (both in same coaching session)
  2. User A creates an overarching goal → User B should see it appear automatically
  3. User A updates an action → User B should see the update without refresh
  4. User A creates an agreement → User B should see it immediately
  5. Verify no sensitive data appears in browser console (F12 → Console)

  **Connection Stability:**
  1. Open browser DevTools → Network tab
  2. Filter for EventSource connections
  3. Verify single SSE connection remains open (not repeatedly closing/reconnecting)
  4. Navigate between pages → connection should persist

  **Backend Logs:**
  * Should see "Sent [event_type] event to N user(s)" 
  * Should NOT see "channel closed" warnings during normal operation

  ### Concerns

  * SWR cache invalidation is intentionally broad (all caches with `backendServiceURL`) for simplicity - could be optimized to specific 
  endpoints if performance becomes an issue
  * SSE connections are ephemeral - offline users miss events (by design, they see fresh data on next page load)
  * Using `provider: () => new Map()` in SWRConfig creates isolated cache per provider instance - ensures clean state but breaks cross-tab
   cache sharing

### DIFF
--- a/src/lib/hooks/use-sse-connection.ts
+++ b/src/lib/hooks/use-sse-connection.ts
@@ -1,20 +1,22 @@
 "use client";
 
-import { useEffect, useRef } from 'react';
+import { useEffect, useState } from 'react';
 import { siteConfig } from '@/site.config';
 import { useSSEConnectionStore } from '@/lib/contexts/sse-connection-context';
 import { logoutCleanupRegistry } from '@/lib/hooks/logout-cleanup-registry';
 
 export function useSSEConnection(isLoggedIn: boolean) {
-  const eventSourceRef = useRef<EventSource | null>(null);
+  const [eventSource, setEventSource] = useState<EventSource | null>(null);
 
   // Get store instance directly - Zustand actions are stable and don't need to be in dependencies
   const store = useSSEConnectionStore((state) => state);
 
   useEffect(() => {
     if (!isLoggedIn) {
-      eventSourceRef.current?.close();
-      eventSourceRef.current = null;
+      setEventSource((prev) => {
+        prev?.close();
+        return null;
+      });
       store.setDisconnected();
       return;
     }
@@ -45,7 +47,7 @@ export function useSSEConnection(isLoggedIn: boolean) {
       }
     };
 
-    eventSourceRef.current = source;
+    setEventSource(source);
 
     const unregisterCleanup = logoutCleanupRegistry.register(() => {
       console.log('[SSE] Cleaning up connection on logout');
@@ -61,5 +63,5 @@ export function useSSEConnection(isLoggedIn: boolean) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [isLoggedIn]);
 
-  return eventSourceRef.current;
+  return eventSource;
 }


src/lib/providers/sse-provider.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const SseProvider = ({ children }: SseProviderProps) => {
+  const [store] = useState(() => createSseConnectionStore());
+
+  return (
+    <SseConnectionStoreContext.Provider value={store}>
+      <SseConnectionManager />
+      {children}
+    </SseConnectionStoreContext.Provider>
+  );
+};
```
jhodapp: Please rename all Sse names instances to SSE to follow other acronym letter capitalization patterns like SWRConfig.

src/lib/contexts/sse-connection-context.tsx
```diff
@@ -0,0 +1,20 @@
+"use client";
+
+import { createContext, useContext } from 'react';
+import { type StoreApi, useStore } from 'zustand';
+import { type SseConnectionStore } from '@/lib/stores/sse-connection-store';
+import { useShallow } from 'zustand/shallow';
+
+export const SseConnectionStoreContext = createContext<StoreApi<SseConnectionStore> | null>(null);
+
+export const useSseConnectionStore = <T,>(
```
jhodapp: Please rename to useSSEConnectionStore.

src/lib/hooks/use-sse-connection.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (!isLoggedIn) {
+      eventSourceRef.current?.close();
+      eventSourceRef.current = null;
+      store.setDisconnected();
+      return;
+    }
+
+    store.setConnecting();
+
+    const source = new EventSource(`${siteConfig.env.backendServiceURL}/sse`, {
```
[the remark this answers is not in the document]
jhodapp: I'd love to see some integration tests that can validate the lifetime correctness of these things, like the lifetime of a single `EventSource` instance and its events like `onopen` and `onerror`.

Especially before we start using this all over the place since we could run into more obscure and tricky bugs to trace down if we don't ensure this is very solid in this single simple use case with overarching goals.

src/lib/stores/sse-connection-store.ts
jhodapp: So the purpose of this is to create a single global SSE connection state manager instance that gets shared? And this is in-memory only, correct? I don't see any persistence of the store to local storage.

src/test-utils/setup.ts
jhodapp: Looks like you were anticipating adding some tests but didn't quite get to it?

src/types/sse-events.ts
jhodapp: Not that these events aren't obvious what they do already, but can you set a pattern of documenting each event type with a code comment above each explaining its purpose and an example of when it gets used? Claude should be able to write these comments very well.

jhodapp (changes requested):
Looking great Caleb, thanks. I've got a fresh round of suggestions and questions for you on this PR.

src/lib/hooks/use-sse-cache-invalidation.ts
```diff
@@ -0,0 +1,70 @@
...
+        // Handle string keys
+        if (typeof key === 'string' && key.includes(baseUrl)) return true;
+        // Handle array keys (SWR supports both formats)
+        if (Array.isArray(key) && key[0] && key[0].includes(baseUrl)) return true;
+        return false;
+      },
+      undefined,
+      { revalidate: true }
+    );
+  }, [mutate, baseUrl]);
```
jhodapp: I'd like to see SWR cache invalidation sharpened up. When we currently receive an SSE event that causes an invalidation, every component on the coaching session page re-renders and this is kind of a jarring user experience. I asked Claude about this and here's what it recommends:

**Root Cause: Overly Broad Cache Invalidation**

The problem is in use-sse-cache-invalidation.ts (lines 12-24).

When any SSE event (including overarching_goal_updated) is received, it invalidates ALL SWR caches that contain the backend URL. This triggers refetches for:

``` ┌────────────────────────────────┬─────────────────────────────┬─────────────────────────────┐
│              Hook              │          Cache Key          │         Re-renders          │
├────────────────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ useCurrentCoachingSession      │ /coaching_sessions/...      │ Page header, title          │
├────────────────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ useCurrentCoachingRelationship │ /coaching_relationships/... │ Session selector, title     │
├────────────────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ useOverarchingGoalBySession    │ /overarching_goals?...      │ Goal component ✓ (intended) │
├────────────────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ Actions hooks                  │ /actions?...                │ Actions tab                 │
├────────────────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ Agreements hooks               │ /agreements?...             │ Agreements tab              │
├────────────────────────────────┼─────────────────────────────┼─────────────────────────────┤
│ Notes hooks                    │ /notes/...                  │ Notes tab                   │
└────────────────────────────────┴─────────────────────────────┴─────────────────────────────┘
```
All these components re-render even though only the overarching goal changed.

**A Way to Fix This Issue**

Scope cache invalidation to only the relevant endpoint:

```typescript
useSseEventHandler(eventSource, 'overarching_goal_updated', () => {
  mutate(
    (key) => {
      const keyStr = typeof key === 'string' ? key : key?.[0];
      return keyStr?.includes('/overarching_goals') ?? false;
    },
    undefined,
    { revalidate: true }
  );
  console.log('[SSE] Revalidated overarching_goals cache');
});
```

src/lib/hooks/use-sse-system-events.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { useAuthStore } from '@/lib/providers/auth-store-provider';
+import { useLogoutUser } from '@/lib/hooks/use-logout-user';
+import { useSseEventHandler } from './use-sse-event-handler';
+
+export function useSseSystemEvents(eventSource: EventSource | null) {
+  const logout = useLogoutUser();
+  const isLoggedIn = useAuthStore((store) => store.isLoggedIn);
+
+  useSseEventHandler(eventSource, 'force_logout', async (event) => {
```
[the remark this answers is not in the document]
calebbourg: On the frontend it exists but not yet on the backend

src/lib/hooks/use-sse-connection.ts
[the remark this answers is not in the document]
calebbourg: Yes you're correct except that the connection is maintained at the root layer so it persists across page navigation, re-renders, etc.

PR #379, opened 2026-04-24
ci: replace PR preview dispatch dropdowns with runtime ref resolver

### LINKED ISSUE #378: CI: replace choice-based PR preview dispatch with runtime ref resolver (stop all bot commits to main)
## Problem

Recent `main` history in this repo includes `github-actions[bot]` commits titled `chore: refresh preview commit choices [skip ci]` (3 of the last 30 at the time of filing). Every PR lifecycle event (`opened` / `reopened` / `closed`) fires `.github/workflows/refresh-preview-commits.yml`, which rewrites `dispatch-pr-preview-frontend.yml`'s `options:` array via `yq` and pushes to `main` with `GHCR_PAT`.

The bot commits pollute history, invalidate `git blame`, and falsely correlate contributors' PR-branch pushes with `main` mutations.

## Root cause

`type: choice` inputs on a `workflow_dispatch` trigger are rendered server-side from the YAML on the default branch. GitHub provides no hook to populate `choices` dynamically at dispatch-render time. The only way to keep the list fresh is to edit the YAML → commit → push. Therefore, `type: choice` is fundamentally incompatible with "zero bot commits to main."

## Fix (mirror of backend sibling issue)

Sibling: refactor-group/refactor-platform-rs#290

1. Replace the `type: choice` inputs on `dispatch-pr-preview-frontend.yml` with:
    - `frontend_pr_number` — required string (e.g. `373` or `PR#373`).
    - `frontend_sha_override` — optional, override the PR branch HEAD.
    - `backend_ref` — optional, default `main`; accepts `main`, a branch name, a SHA, or `PR#<num>`.
2. Add a resolver step at the top of the validate job that resolves each input to a concrete SHA + branch using the GitHub API.
3. **Delete `.github/workflows/refresh-preview-commits.yml`** in this repo too.
4. The new `list-preview-refs.yml` lives in the backend repo (single source of truth) and can be dispatched from the GitHub UI there.
5. Update `.claude/CLAUDE.md` "Commit Choice Refresh" section to describe the new pattern.

## Non-regressions

- Deploying the HEAD of a frontend PR against backend main → pass `frontend_pr_number=<num>`, leave `backend_ref=main` (default).
- Deploying an open frontend PR against an open backend PR → pass `backend_ref=PR#<num>`.
- The reusable `ci-deploy-pr-preview.yml@main` in the backend repo is unchanged — still called with `backend_sha` / `frontend_sha` from the resolver outputs.

## Gotchas / trade-offs

- UX shift: dropdown → text input. Users run `List Preview Refs` (a helper workflow in the backend repo) or look at the PR URL.
- `GHCR_PAT` still required where the workflow modifies nothing — the delete of `refresh-preview-commits.yml` removes the only writer to `.github/workflows/` from this repo.

## Expected outcome

- Zero `github-actions[bot]` commits on `main` going forward.
- PR preview deploys continue to work end-to-end with unchanged behavior for users.

lmcdonough:
Closes #378. Mirrors refactor-group/refactor-platform-rs#291.

## Context

`main` in this repo has been receiving `chore: refresh preview commit choices [skip ci]` commits from `github-actions[bot]` on every PR lifecycle event (`opened` / `reopened` / `closed`). That pollutes history, invalidates `git blame`, and falsely correlates contributors' PR-branch pushes with `main` mutations.

## Root cause

`type: choice` inputs on a `workflow_dispatch` trigger are rendered server-side from the YAML on the default branch. GitHub provides no hook to populate `choices` dynamically at dispatch-render time. The only way to keep the list fresh is to edit the YAML → commit → push. `type: choice` is fundamentally incompatible with "zero bot commits to main."

## Fix

- **Replace dropdowns with three text inputs** on `dispatch-pr-preview-frontend.yml`:
  - `frontend_pr_number` — required (accepts `373` or `PR#373`).
  - `frontend_sha_override` — optional, override the PR branch HEAD.
  - `backend_ref` — optional, default `main`; accepts `main`, a branch name, a 7+ char SHA, or `PR#<num>`.
- **Resolve inputs at dispatch time** via `gh pr view` / `gh api repos/.../commits/<ref>`. Nothing is written to `main`.
- **Delete `refresh-preview-commits.yml`** — it exists only to keep the dropdowns fresh.
- **Update `.claude/CLAUDE.md`** to document the new flow and point to the `list-preview-refs.yml` helper in the backend repo.

Diffstat: `3 files changed, 112 insertions(+), 258 deletions(-)` — net code reduction.

The backend PR adds `list-preview-refs.yml` — a read-only helper workflow that prints current `main` commits + open PRs from both repos to the run summary. That's a single source of truth; there's no need to duplicate it here.

## Non-regressions

| Capability | How it still works |
|---|---|
| Deploy frontend PR against `main` | `frontend_pr_number=<num>`, `backend_ref=main` (default) |
| Deploy frontend PR against a backend PR | add `backend_ref=PR#<num>` |
| Deploy specific SHAs | `frontend_sha_override=<sha>` / `backend_ref=<sha>` |
| Draft PRs | `PR_STATE != OPEN` gate still allows drafts |
| Cross-repo pair selection | `backend_ref` still queries the backend repo |
| Cleanup on PR close | Unchanged |
| Reusable `ci-deploy-pr-preview.yml@main` | Unchanged — still called with `backend_sha` / `frontend_sha` / `pr_number` / `branch_name` |

## Test plan

Already validated locally:
- `actionlint .github/workflows/dispatch-pr-preview-frontend.yml` → clean.
- `act pull_request -l` → no match for `refresh-preview-commits` or `Refresh Preview Commits` (proof that PR events can no longer produce a main commit from these workflows).
- `act workflow_dispatch -W dispatch-pr-preview-frontend.yml --input frontend_pr_number=373 --input backend_ref=main -n` → dry-run succeeds end-to-end.

Post-merge GitHub-side smoke:
1. `gh workflow run dispatch-pr-preview-frontend.yml -f frontend_pr_number=<live-PR> -f backend_ref=main` → confirm deploy end-to-end.
2. `gh run list --workflow refresh-preview-commits.yml` → empty (file is gone).
3. 48h observation: `git log --author='github-actions\[bot\]' main --since=<merge-date>` → empty.

## Gotchas

- Land both PRs (this one + refactor-group/refactor-platform-rs#291). Each repo removes its own `refresh-preview-commits.yml`; leaving one behind would keep polluting that repo's `main`.
- `GHCR_PAT` is no longer needed by the modified dispatch workflow — it uses `GITHUB_TOKEN` for API reads.

## Note on PR re-open (attempted GH Actions workaround)

I closed and reopened this PR on 2026-05-01 to try to force GitHub Actions to re-register the new `.github/workflows/test-dispatch-regex.yml` workflow against this PR. After multiple pushes (`98c84ff`, `430963a`, `6c8052a`, `1a12c9a`) the `pull_request` trigger has refused to fire on the head branch even though the close/reopen events did register on GitHub's side. The backend twin (refactor-group/refactor-platform-rs#291) triggered cleanly on identical content, so this is a per-branch GH internal state issue, not a workflow bug.

Verification path that does work:
1. `bats .github/workflows/tests/dispatch-pr-preview-frontend.bats` locally: 50/50 pass.
2. `act pull_request -W .github/workflows/test-dispatch-regex.yml -j bats` locally with the same `ubuntu-latest` runner image: 50/50 pass.
3. Backend repo's identical workflow run (`refactor-platform-rs` Test Dispatch Regex on 2026-05-01): 50/50 pass.
4. Frontend repo's `Build, Test & Push to GHCR` was kicked manually via `workflow_dispatch` on this branch and the new `dispatch-pr-preview-frontend.yml` content went through Lint, Build & Test, and Docker push successfully.

After this PR merges to main, GitHub will register the workflow normally and the `pull_request` triggers will start firing on subsequent PRs as expected.


.github/workflows/dispatch-pr-preview-frontend.yml
```diff
@@ -67,121 +55,117 @@ permissions:
...
+          BE_REPO="$OWNER/refactor-platform-rs"
+          if [[ "$INPUT_BACKEND_REF" =~ ^PR#([0-9]+)$ ]]; then
+              BE_PR="${BASH_REMATCH[1]}"
+              BACKEND_FULL_SHA=$(gh pr view "$BE_PR" --repo "$BE_REPO" \
+                  --json headRefOid --jq '.headRefOid' 2>&1) || {
+                  echo "::error::Backend PR #$BE_PR not found in $BE_REPO"
+                  exit 1
+              }
+              echo "::notice::Backend ref resolved from PR #$BE_PR: ${BACKEND_FULL_SHA:0:7}"
+          elif [[ "$INPUT_BACKEND_REF" =~ ^[0-9a-fA-F]{7,40}$ ]]; then
```
jhodapp: Edge case: a branch named purely in 7–40 hex chars (e.g. `abc1234`) gets classified as a SHA here. Rare in practice, and `gh api .../commits/<ref>` resolves either kind — so the only visible effect is the `::notice::` wording. A one-line comment noting the precedence (PR# → SHA → branch) would help future readers.

.github/workflows/dispatch-pr-preview-frontend.yml
```diff
@@ -67,121 +55,117 @@ permissions:
...
+              echo "::notice::Backend ref resolved from SHA: ${BACKEND_FULL_SHA:0:7}"
+          else
+              # Treat as branch name (default path — includes "main")
+              BACKEND_FULL_SHA=$(gh api "repos/$BE_REPO/commits/$INPUT_BACKEND_REF" \
+                  --jq '.sha' 2>&1) || {
+                  echo "::error::Backend branch/ref '$INPUT_BACKEND_REF' not found in $BE_REPO"
+                  exit 1
+              }
+              echo "::notice::Backend ref resolved from branch '$INPUT_BACKEND_REF': ${BACKEND_FULL_SHA:0:7}"
           fi
```
jhodapp: The SHA branch and the branch-name branch both call `gh api repos/$BE_REPO/commits/$INPUT_BACKEND_REF` — they differ only in error message and log line. Could collapse into a single `else`, or keep split for clearer notices. Trade-off either way; flagging in case you prefer the tighter form.

jhodapp (commented):
Looking @lmcdonough. I've included the review I did with Claude.

Reminder: this PR is coupled to refactor-group/refactor-platform-rs#291. Both must land — leaving the backend's `refresh-preview-commits.yml` in place would keep polluting that repo's `main`. Worth calling out in the merge order.

.github/workflows/dispatch-pr-preview-frontend.yml
lmcdonough: Addressed in 98c84ff. I added a precedence comment to the resolver block (PR# wins, then everything else falls through to gh api commits/<ref>) so the intentional behavior for hex-named branches is documented at the call site. I also collapsed the two arms (see thread on line 158), which makes the visible difference between SHA-classified and branch-classified inputs go away entirely. Bats coverage in .github/workflows/tests/dispatch-pr-preview-frontend.bats pins the precedence (50 cases, including the hex-named-branch edge case).

.github/workflows/dispatch-pr-preview-frontend.yml
lmcdonough: Addressed in 98c84ff. I went with the tighter form. Both arms called gh api repos/$BE_REPO/commits/$INPUT_BACKEND_REF and only differed in the notice wording, so collapsing them keeps the resolver honest about what it actually does. Single ::notice:: now reports the resolved short SHA without claiming to know whether the input was a SHA or a branch. Tests cover both shapes (hex-only and named branches) so a future regression flips a bats case rather than silently changing log text.

lmcdonough:
Thanks for the review @jhodapp. Addressed in 98c84ff.

**Coupling and merge order.** Acknowledged. I will land refactor-group/refactor-platform-rs#291 first so refresh-preview-commits.yml is removed before this one merges. Otherwise the backend keeps polluting main with the chore commits this work is meant to eliminate.

**Inline threads.** Both resolved. I collapsed the SHA and branch arms into a single fall-through (the commits/<ref> endpoint is ref-type-agnostic) and added a precedence comment so the hex-named-branch edge case is documented.

**Tests.** Added a 50-case bats suite at .github/workflows/tests/dispatch-pr-preview-frontend.bats covering the PR# / numeric / SHA / branch / override classification logic. The test-dispatch-regex.yml workflow runs it on every PR that touches the dispatch workflow or the bats file, so a regex regression now fails review rather than dispatch.

.github/workflows/test-dispatch-regex.yml
jhodapp, line 1: @lmcdonough Just a heads up: adding `test-dispatch-regex.yml` means another entry in the repo's Actions sidebar (it'll show up as "Test Dispatch Regex"). GitHub's sidebar is file-based, there's no hidden: flag, so every workflow file gets its own row whether or not it's meant for humans to click.

A few options if we want to keep the list tight:

1. Consolidate — fold the bats job into `dispatch-pr-preview-frontend.yml` as an additional job gated by paths:. One file, one sidebar entry, two jobs. Same trick could apply to `cleanup-pr-preview-frontend.yml`.
2. Drop `workflow_dispatch`: on workflows that aren't meant to be human-triggered, doesn't remove the entry but kills the "Run workflow" button so nobody fires it by accident.
3. `on: workflow_call:` only for true child workflows, still appears in the sidebar but stays visually quiet since it only runs when called.

jhodapp (changes requested):
@lmcdonough Let me know if you agree with this or not. I'm finding it harder to find the job that I want to run in the list of GH Actions in both repos, so I'd really like this keep this list as short as possible as consolidate the rest as jobs in a few workflow files.

.github/workflows/test-dispatch-regex.yml
lmcdonough: Good catch — addressed in b5bda75. Removed `test-dispatch-regex.yml` and folded the bats suite into `build_and_push_nonproduction_images.yml` as a parallel `Dispatch Regex Tests` job, so the sidebar stays uncluttered while regex regressions still get caught at PR review time.

Also extended `dispatch-pr-preview-frontend.bats` with 23 new cases for the `validate_input` field-protection helper added in 4485443 (length cap to 250 chars, no newline/CR/tab, charset restriction to `[A-Za-z0-9._/#+@-]` — rejects shell metacharacters, spaces, quotes, parens, etc.). All 73 cases pass locally.

jhodapp (approved):
Looks awesome @lmcdonough, thanks for seeing this through to an excellent place!

Feel free to merge when ready.

PR #390, opened 2026-05-12
Implements creation recurring sessions

### LINKED ISSUE #345: Schedule coaching session series
As a coach, I want to be able to schedule a recurring series of coaching sessions so I don't have to schedule one-off sessions each week.

Other related functionality needed:
1. Set session duration (this is probably needed as its own focused PR on the backend first) (also note that the frontend has a notion of session duration already, but it's hardcoded but does have a function wrapping it providing a lightweight interface already)
2. Set repetitive frequency of sessions
3. Set a start date and end date

Design consideration for on-demand TipTap notes creation:
1. Only create a new TipTap doc **if and only if** the coach or coachee in that CoachingRelationship tries to fetch that coaching session for the first time

rnambaale:
## Description

Adds support for coaches to schedule recurring coaching sessions from the dashboard's "Add new" flow in one shot, and smooths the brief gap where a freshly-created session's TipTap collaboration document is still being provisioned server-side when the user first joins it.

The recurring create flow is wired to the new backend route `POST /coaching_sessions/recurring`. The form enforces the backend's field rules client-side so users get inline feedback instead of 422s.

### Changes

* New `src/types/recurrence.ts` module defining `Frequency`, `Weekday`, `Recurrence`, and `CreateRecurringSessionRequest` types. The end-condition is modeled as a `RecurrenceEnd` discriminated union (`{kind:"count"} | {kind:"until"}`) so the "exactly one of count/until" backend rule is unrepresentable to violate.
* `CoachingSessionApi.createRecurring()` hits `POST /coaching_sessions/recurring` and returns the created sessions.
* `CoachingSessionForm` gains a **Repeats** switch (create mode only). When toggled on it reveals:
  * Frequency select (Daily / Weekly / Bi-weekly / Monthly).
  * `Every N` interval input.
  * Mon–Sun weekday toggle group (only for Weekly / Bi-weekly).
  * `Ends` radio group: `After N occurrences` vs `On <date>`.
* Client-side guards mirroring the backend's field rules: `interval >= 1`, exactly one end condition, `by_weekdays` must include the first-session weekday on weekly/biweekly, `count <= 365`, span `<= 366` days. A single inline message surfaces the first violation and gates the submit button.
* The submit button branches: create + recurring -> `createRecurring`; create + non-recurring -> existing single-session create; update -> unchanged.
* The editor loading state in `CoachingNotes` is reframed as **"Preparing your coaching notes..."** with a delayed reassurance line after 1.5s ("New sessions can take a few seconds to get ready the first time."). Suppresses the hint on the fast path while making the slow path feel intentional.
* `useCollaborationToken` retries on transient failures with exponential backoff (`300ms -> 4.8s`, capped at 5 attempts, ~9s total — fits inside the 10s editor sync timeout). Per-attempt errors are no longer surfaced to callers; `isError` is only set when retries are genuinely exhausted, so the editor stops flashing its error UI during the brief window when the TipTap doc is being created.

### Screenshots / Videos Showing UI Changes

Screenshot 2026-05-12 at 3 16 43 PM

---

Screenshot 2026-05-12 at 3 16 28 PM

### Testing Strategy

* **Recurring create (happy path):** Open the create-session dialog as a coach, pick a coachee and a date/time, toggle **Repeats**, leave Weekly default, change `Ends` to `On <date>` ~3 weeks out, submit. Confirm:
  * Sessions appear in the dashboard immediately (refresh fires after the dialog closes).
  * Opening the first session loads notes cleanly without an error flash.
  * Opening a later session (still freshly-created) shows "Preparing your coaching notes..." briefly, then the editor loads.
* **Recurring create (validation):**
  * Pick a Wednesday `start_at`, switch to weekly, deselect Wed -> submit button disabled with message naming Wed.
  * Choose `End on <date>` more than 366 days out -> button disabled.
  * Choose `After N occurrences` with N=400 -> button disabled.
  * Daily frequency hides the weekday toggle group; Monthly likewise.
* **Backend rule enforcement at the edge:** Temporarily relax a client-side guard to confirm a 422 surfaces the toast: "Recurrence is invalid — check weekdays, end condition, and the 365 occurrence / 366 day caps."
* **Non-recurring create / update:** Confirm the existing single-session create and the update flows are unchanged.
* **Editor join smoothing:** With the backend introducing artificial latency on `GET /coaching_sessions/{id}` (or by joining a session immediately after creating it), confirm the editor stays on the loading state and never flashes the "Could not load coaching notes" error before the JWT eventually resolves.

### Concerns

* **Update flow doesn't support editing the recurrence rule** — only the single occurrence. We may want a follow-up to support "edit this and all following" semantics.


src/components/ui/dashboard/coaching-session-form.tsx
```diff
@@ -186,17 +268,69 @@ export default function CoachingSessionForm({
...
+        if (until < start) return "End date must be on or after the first session.";
+        const spanDays = Math.floor(until.diff(start, "days").days) + 1;
+        if (spanDays > MAX_SPAN_DAYS) {
+          return `End date is beyond the ${MAX_SPAN_DAYS}-day window.`;
+        }
+      }
+    }
+    if (frequencySupportsWeekdays(frequency)) {
+      if (byWeekdays.length === 0) {
+        return "Pick at least one weekday.";
```
jhodapp: This branch is effectively unreachable: the auto-seed useEffect (line 152) re-seeds byWeekdays with startWeekday whenever the array is empty, so a user can't actually get to a zero-weekday state. Either drop this validation branch or reconsider whether the effect should keep fighting the user's deselection.

src/components/ui/dashboard/coaching-session-form.tsx
```diff
@@ -249,6 +383,153 @@ export default function CoachingSessionForm({
...
+                      </SelectContent>
+                    </Select>
+                  </div>
+                  <div className="space-y-2">
+                    <Label htmlFor="recurrence-interval">Every</Label>
+                    <Input
+                      id="recurrence-interval"
+                      type="number"
+                      min={1}
+                      max={MAX_SPAN_DAYS}
```
jhodapp: `max={MAX_SPAN_DAYS}` (366) is semantically off — interval is a step multiplier ("every N weeks"), not a day count. Suggest a smaller, meaningful cap (e.g. ~52) or no max.

src/lib/api/collaboration-token.ts
```diff
@@ -26,17 +36,47 @@ export const useCollaborationToken = (coachingSessionId: string) => {
...
     revalidateOnReconnect: false,
+    onSuccess: () => setRetriesExhausted(false),
+    onErrorRetry: (err, _key, _config, revalidate, { retryCount }) => {
+      const status = (err as { response?: { status?: number } })?.response
+        ?.status;
+      if (status === 401 || status === 403) {
+        setRetriesExhausted(true);
+        return;
+      }
+      if (retryCount >= MAX_TOKEN_RETRIES) {
```
jhodapp: The comment at line 17–22 claims "5 attempts (300ms → 4.8s)", but with `retryCount >= MAX_TOKEN_RETRIES` as the stop and delay BASE * 2 ** retryCount, the actual count/delays depend on whether SWR's retryCount starts at 0 or 1. If it starts at 1, the first delay is 600ms and you get 4 retries, not 5 — which would change the "~9s, inside the 10s sync timeout" guarantee. Worth confirming.

src/lib/api/collaboration-token.ts
```diff
@@ -26,17 +36,47 @@ export const useCollaborationToken = (coachingSessionId: string) => {
...
+        setRetriesExhausted(true);
+        return;
+      }
+      setTimeout(
+        () => revalidate({ retryCount }),
+        TOKEN_RETRY_BASE_MS * 2 ** retryCount
+      );
+    },
```
jhodapp: Retry/backoff in useCollaborationToken is hand-rolled, and there's no existing retry utility or shared SWR config to reuse, so it's not duplication today. But it's the kind of logic that tends to get copy-pasted later. If a second hook ever needs backoff, this should be extracted to a shared SWR config/helper. What do you think?

jhodapp (changes requested):
Thanks for this Raymond, nice work.

A few things to mention in here about the functionality:
1. It's missing a toast message after successfully adding the reoccurring sessions.
2. It should say "days" instead of "weekdays" since it includes weekend days as well

Image

3. This should use the same calendar date picker used in other parts of the application. This one I think is the default radix one but I believe the others are using the one in `src/components/ui/calendar.tsx`

Image

4. When I clicked on the "Ends" "on" I got this error right away even though I wasn't given a chance to enter a date yet.

Image

5. The way you've architected this in the backend, even though I know we discussed having a cutoff date no matter what, is it possible to have a "never" option? I'm curious and thought I'd ask again.

6. No automated tests in this PR. recurrence.ts has pure, highly testable functions (recurrenceToPayload, weekdayFromLuxon, frequencySupportsWeekdays) and the recurrenceError validation has dense branching (caps, span math, weekday inclusion) that's easy to regress. Could you add at least unit coverage for recurrence.ts and the retry/terminal-error logic in useCollaborationToken before merge?

src/components/ui/dashboard/coaching-session-form.tsx
```diff
@@ -177,7 +257,9 @@ export default function CoachingSessionForm({
           ? "Could not connect to server. Please check your internet connection."
           : error.status === 502
             ? "Could not create Google Meet link due to a connection error. Please try again."
-            : `Failed to ${mode} coaching session. Please try again.`;
+            : error.status === 422 && isRecurring
+              ? "Recurrence is invalid — check weekdays, end condition, and the 365 occurrence / 366 day caps."
+              : `Failed to ${mode} coaching session. Please try again.`;
```
jhodapp, line 292: `error.status === 422 && isRecurring` assumes a 422 is always about the recurrence rule, but it could also be an invalid coaching_relationship_id or start_at. The "check weekdays, end condition, and the 365/366 caps" message could misdirect the user in those cases. Low priority, but consider a more general fallback.

jhodapp (changes requested):

src/components/ui/dashboard/coaching-session-form.tsx
rnambaale: I have created a separate `MAX_INTERVAL` guardrail for this. 3a996d8499d71d493f52d5adc7f52984148cfa3d

src/components/ui/dashboard/coaching-session-form.tsx
rnambaale: I think now that we have extracted the validation to a helper outside the component, it might actually be possible for a caller to reach the branch.

src/lib/api/collaboration-token.ts
rnambaale: Good catch. I've confirmed SWR's `retryCount` is 1-based.

The retries now match the list in the comment (300, 600, 1.2s, 2.4s, 4.8s ≈ 9.3s. 73da03df9c5e67654d92f89b332e89ad58d190e5

src/lib/api/collaboration-token.ts
rnambaale: Agreed, I think we shall extract it when it is needed for the second caller.

src/components/ui/dashboard/coaching-session-form.tsx
rnambaale: I have implemented a more generic fallback here: 257c5a1269a7e28023f2ce0b7af7e65580bc1276

src/lib/api/collaboration-token.ts
jhodapp: @rnambaale would you mind extracting today it even though it's only used in one place, this is just the kind of thing that gets skipped over by LLMs. Another option is just adding a clear TODO comment above this section giving instructions on the plan to extract it into a reusable piece method/hook.

src/components/ui/dashboard/coaching-session-form.tsx
jhodapp: Ok, sounds good.

jhodapp (changes requested):

src/components/ui/dashboard/coaching-session-form.tsx
```diff
@@ -249,6 +395,197 @@ export default function CoachingSessionForm({
...
+                          aria-label={weekdayLabel(d)}
+                          className="w-10"
+                        >
+                          {weekdayLabel(d).slice(0, 1)}
+                        </ToggleGroupItem>
+                      ))}
+                    </ToggleGroup>
+                    {startWeekday && (
+                      <p className="text-xs text-muted-foreground">
+                        First session is a {weekdayLabel(startWeekday)} — it must be selected.
```
jhodapp: @rnambaale I'm not sure what this message is telling me when it appears in the form.

rnambaale: So what that message says is: 
"When `by_weekdays` is sent for a `weekly`/`biweekly` recurrence; the day where the selected `start_at` lies must be selected."

Screenshot 2026-05-19 at 4 28 42 PM

src/lib/api/collaboration-token.ts
rnambaale: Extacted: 53521baf09d1509d35275a5896d7dddd2e1d024e

src/components/ui/dashboard/coaching-session-form.tsx
jhodapp: @rnambaale ok, that's still not very clear to me. I'm wondering if we can improve this wording, and also steer away from using em dashes (i.e. --).

I'm still not sure I fully understand it otherwise I'd make a suggestion here.

rnambaale: @jhodapp Do you think this is a better version?  

I have modified it to read like:

`Your first session is on a Mon, which must stay selected.`

jhodapp: > @jhodapp Do you think this is a better version?
> 
> I have modified it to read like:
> 
> `Your first session is on a Mon, which must stay selected.`

@rnambaale I think this is reasonable and makes more sense. Let's go with it. Thanks!

jhodapp (approved):
Approved. I'm happy to merge and deploy this new feature for you after my work day tomorrow if you're ready for that.

